PR #76, opened 2018-11-24
Fixed help message

### LINKED ISSUE #70: Styrange usage()
```
-y --wipe-db                    Do not wipe the DB during parse stage
```

I suspect that what is meant is:

```
-y --wipe-db                    Wipe the DB during parse stage
```

All in one this is not clear what does that mean. Which DB? What is the impact for the user to choose this option?

jtaylor351:
Tried to resolve #70 by adding the correct help message and making the message more descriptive. Also, I fixed the logging related to --wipe-db flag


dattaz:
Thanks you for your PR !
I'm OK with your PR. See my comment in #70

PR #103, opened 2020-04-23
JSBeautify Formatting

aschlumpf:
The purpose of this PR is to bring in JSBeautify -- a JavaScript code formatter. The group has applied this formatting to our other PR already, so if this PR is merged first the diff will be a bit easier to digest. I did not install JSBeautify locally, as the project did not already have a `package.json` file. Instead, I installed the package globally to my environment using `npm install -g js-beautify`. Once installed, the code can be formatted by entering the command `js-beautify path/to/js/files --django`. The `--django` flag respects the templating that exists in the JS files. 

Note that I could not use Prettier, a more powerful formatting tool which actually can modify JavaScript AST, because of the templating. JSBeautify should be sufficient for this use case of handling PR's, though.


rgaudin:
why was this closed? will you submit another one?

aschlumpf:
Yes I will re-open soon. I closed because we saw that some of the string changes I did by hand caused syntax errors (I did a CTRL + F to convert single quotes to double quotes). I want to build a zim before submitting the PR.

aschlumpf:
Everything built on my local... re-opening.

rgaudin:
OK, I've tested it.

What you didn't say is that it should be invoked as 

```
js-beautify --config gutenbergtozim/.jsbeautifyrc --templating django gutenbergtozim/templates/js/l10n.js gutenbergtozim/templates/js/tools.js
```

Using this I could confirm your change to `l10n.js` but not `tools.js` which is completely different. I believe you used different settings.

aschlumpf:
There may be some differences because I manually changed some strings from using single-quotes to double-quotes. The reason I did this was because our main PR used a different formatter which was able to modify the AST and change strings. 

However, that formatter cannot run automated as the jinja templating makes it fail parsing. In order to try and resolve some conflicts, I tried to manually align these two branches in regard to string convention. That shouldn't be necessary going forward, but it may make the diff for out PR a bit easier to read.

In regards to the command-line, yes you must specify the directory. When I called the command I was in the repo so I didn't need the full path. If you call `js-beautify` from directory where a `.jsbeautifyrc` exists it will automatically respect it. I also just added the django setting to be on by default.

PR #117, opened 2020-06-18
Fixes #115 - Better temporary files handling

### LINKED ISSUE #115: gutenberg stores data outside output folder
(At least) when using `--one-language-one-zim`, the folder specified is not used for temporary data storage.

It should be fixed the way we're doing it for youtube and ted as this can be dozens of GB outside the requested location.

satyamtg:
Fixes #115 by introducing following changes -
- static folder is now a temporary directory unless specified
- introduced --tmp-dir to specify where to store static folder (if not specified) and other temporary files
- "tmp" directory is now a real temporary directory
- fixed leftovers when optimizing epub

The deletion of static folder is kept intact (delete if ZIM is being made, else keep). The newly added --tmp-dir argument takes the folder to store all temporary files and the static folder (if not specified by --static-folder). This allows us to retain the previous behaviour (i.e. running exploded parts of the process individually)

### DIFF
--- a/gutenberg2zim
+++ b/gutenberg2zim
@@ -9,6 +9,8 @@ import sys
 from docopt import docopt
 from path import Path as path
 import shutil
+import pathlib
+import tempfile
 
 from gutenbergtozim import logger, VERSION
 from gutenbergtozim.database import setup_database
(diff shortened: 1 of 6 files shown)


rgaudin:
Honestly I don't think we need this.

There is no *output folder* in the current implementation. If not using the one-lang-one-zim option, you don't specify an output folder and that just goes to CWD.

We have options for specifying some folders and resources but that's not very helpful as cwd will anyway end up the large ZIM files.

I am fixing #115 by changing the Dockerfile to work off the `/output` folder which is configured on the ZIM farm. This way, `/output` is the cwd and everything gets stored there.

We'll change all this once we refactor. I suggest we close this as it doesn't completely fix it and adds a bit more complexity to this fragile edifice :)

* Can you open a separate PR for the epub optim fix?
* Why are you encoding some logs ?

satyamtg:
> * Why are you encoding some logs ?
Oh. I didn't see that. That was a mistake (didn't pull to the temporary folder where I was testing and copied from there) I'll take care that this doesn't happen again. Anyways, Since it's getting closed, it shouldn't matter.

rgaudin:
OK I thought there was a reason for that as I had removed them previously to have better-looking logs. Good then.

PR #127, opened 2020-06-27
Consider having consistent data always

### LINKED ISSUE #126: Gutenberg run failed on zimfarm
The last gutenberg run (https://farm.openzim.org/pipeline/5ef4da41443d22424a730e05/debug) failed possibly due to a connection error while downloading a book. The scraper shouldn't crash on such errors but should definitely output the error.

This is the last log from the scraper explaining the message -
```
requests.exceptions.ConnectTimeout: HTTPConnectionPool(host='aleph.gutenberg.org', port=80): Max retries exceeded with url: /5/2/5/5254/5254-pdf.pdf (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x7fc4d3b67390>, 'Connection to aleph.gutenberg.org timed out. (connect timeout=20)')
```

satyamtg:
This has the following changes -
- commented out resource_exists() and its use while downloading as we have consistent data from rsync step
- edited resource_exists() so that scraper doesn't fail on error with requests in a particular URL
- only send filtered URLs to download function, instead of sending all combinations if filtered list is empty (commented that code as can be used in development scenarios)

This fixes #126

### DIFF
--- a/gutenbergtozim/urls.py
+++ b/gutenbergtozim/urls.py
@@ -125,6 +125,7 @@ def build_urls(files):
                 for u in possible_url
                 if Url.get_or_none(url=urlparse.urlparse(u).path[1:])
             ]
+            # Use only the URLs in DB
             files[i] = filtre
             # for development
             # if len(filtre) == 0 and len(possible_url) != 0:
--- a/gutenbergtozim/urls.py
+++ b/gutenbergtozim/urls.py
@@ -125,10 +125,12 @@ def build_urls(files):
                 for u in possible_url
                 if Url.get_or_none(url=urlparse.urlparse(u).path[1:])
             ]
-            if len(filtre) == 0 and len(possible_url) != 0:
-                files[i] = possible_url
-            else:
-                files[i] = filtre
+            files[i] = filtre
+            # for development
+            # if len(filtre) == 0 and len(possible_url) != 0:
+            #     files[i] = possible_url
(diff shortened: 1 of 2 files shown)


gutenbergtozim/urls.py
```diff
@@ -125,10 +125,12 @@ def build_urls(files):
                 for u in possible_url
                 if Url.get_or_none(url=urlparse.urlparse(u).path[1:])
             ]
-            if len(filtre) == 0 and len(possible_url) != 0:
-                files[i] = possible_url
-            else:
-                files[i] = filtre
+            files[i] = filtre
```
rgaudin, line 128: Add a comment explaining this restricts to use only what we have in DB

satyamtg: Done

satyamtg:
> This is OK and will prevent testing unwanted links but we may want to download those files with a more robust downloader to have extensive retries are we are talking about dozens of thousands of files so the risk for errors (connection mostly) is very high (although mitigated by S3).

The downloads actually occur through curl as seen in https://github.com/openzim/gutenberg/blob/825470482e30bd3ef70e79e587780504c0fc7fa4/gutenbergtozim/utils.py#L101 .
We might want to have it replaced by save_large_file() from scraperlib as that would eventually be maintained in a much better manner there. Opening an issue for this.

PR #133, opened 2020-07-09
Raise error on zimwriterfs failure and update download cache always

satyamtg:
Runs zimwriterfs directly through subprocess, captures its output, and raises SystemExit if returncode isn't 0. Also logs zimwriterfs output in that case.

Also this allows updating download cache with optimized versions of files if S3 URL was not passed. This makes it helpful for future runs on same dl-cache

### DIFF
--- a/gutenbergtozim/zim.py
+++ b/gutenbergtozim/zim.py
@@ -3,11 +3,12 @@
 # vim: ai ts=4 sts=4 et sw=4 nu
 
 import six
+import subprocess
 
 from path import Path as path
 
 from gutenbergtozim import logger, VERSION
-from gutenbergtozim.utils import exec_cmd, get_project_id, FORMAT_MATRIX
+from gutenbergtozim.utils import get_project_id, FORMAT_MATRIX
 from gutenbergtozim.iso639 import ISO_MATRIX
 from gutenbergtozim.export import export_skeleton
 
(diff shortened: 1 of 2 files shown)


gutenbergtozim/zim.py
```diff
@@ -101,7 +102,12 @@ def build_zimfile(
 
     if not create_index:
         cmd.insert(1, "--withoutFTIndex")
-    if exec_cmd(cmd) == 0:
+    zimwriterfs = subprocess.run(
```
rgaudin, line 105: We don't need the pipe ; it's OK (wanted actually) to have the zimwriterfs output printed.

gutenbergtozim/zim.py
```diff
@@ -101,7 +102,12 @@ def build_zimfile(
...
+        cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True
+    )
+    if zimwriterfs.returncode == 0:
         logger.info("Successfuly created ZIM file at {}".format(zim_path))
     else:
         logger.error("Unable to create ZIM file :(")
+        logger.debug(zimwriterfs.stdout)
+        raise SystemExit
```
rgaudin, line 113: we want to forward the exit code of zimwriterfs here. It's important. so `raise SystemExit(zimwriterfs.returncode)`

satyamtg:
> For the upload, what you added is that if running with S3 enabled but you're not downloading a file because it's already present on disk, you'd upload it to the S3 cache ? Is that it?
> Then we don't want that.
> We want to only upload to S3 after optimizing, after downloading from source because it was not in S3.
> This gutenberg-only dl-cache will be removed anyway.

Nope. Previously what happened was this -
If we do not give the --optimization-cache argument while running, downloads and optimizations went well. However, after optimizing, the optimized files were kept only in static. The download cache was not updated to have the optimized files instead of the unoptimized one. So, if you run export steps multiple times, the optimization would take place always.
What happens now is -
If we do not give the --optimization-cache argument while running, downloads and optimizations go well and after the optimization, the dl-cache contains the optimized file in optimized_dir instead of unoptimized files in unoptimized_dir

gutenbergtozim/zim.py
satyamtg: Removed that

gutenbergtozim/zim.py
satyamtg: Done

PR #134, opened 2020-07-11
Small fixes

satyamtg:
This does the following -

- While downloading something with wget, it seems that it creates a dummy file even if it fails to download anything. This file is now cleaned up properly in download_file(). This happened for some covers.

- Added verbose logging to zimwriterfs

- Previously, even if something was downloaded successfully from optimization cache, the scraper went on to check other URLs and download. This is fixed with proper break statements. 

- Books that could not be downloaded in any format are now removed from DB to prevent errors being thrown at the time of exporting.

### DIFF
--- a/gutenbergtozim/download.py
+++ b/gutenbergtozim/download.py
@@ -258,7 +258,7 @@ def download_book(
                         optimizer_version=optimizer_version,
                     ):
                         downloaded_from_cache = True
-                        continue
+                        break
                 if not download_file(url, zpath):
                     logger.error("ZIP file donwload failed: {}".format(zpath))
                     continue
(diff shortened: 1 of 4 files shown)


gutenbergtozim/download.py
```diff
@@ -317,8 +319,15 @@ def download_book(
             # delete instance from DB if download failed
             logger.info("Deleting instance from DB")
             bf.delete_instance()
+            unsuccessful_formats.append(book_format)
             pp(allurls)
 
+    # delete book from DB if not downloaded in any format
+    if all(format in unsuccessful_formats for format in ["html", "pdf", "epub"]):
```
rgaudin, line 326: Don't use `format`, it's a reserved keyword.
Also, this is hardly readable.

```py
if len(unsuccessful_formats) == len(formats):
```

satyamtg: Oh yup. That went out of the mind

PR #140, opened 2021-01-12
Now JS will be added during setup

hksingh4545:
Now JS of DataTables-1.10.13 and Responsive-2.1.1 will be added during setup


kelson42:
@hksingh4545 Thx

rgaudin (changes requested):
Thanks. I see the files are not identical. Diff tells me for example that the new file refers to 

```css
.sorting{background-image:url("/DataTables-1.10.13/images/sort_both.png")
```

Kinda the same for all images.

Now I see that we did not have those images in the repo. I can't say if we are not using the features using them or if it's broken but this is a good opportunity to fix it. If that's not a nightmare to fix ; as it might be difficult to find this old archive (my understanding is that it's a bundle).

Also we should only keep one version (minified?) of CSS and JS.

kelson42:
@hksingh4545 Any feedback?

kelson42:
@hksingh4545 Without feedback, I will have to close this ticket unfortunately :( Would you be able please to answer @rgaudin' comment?

hksingh4545:
can you explain a little more about it?

rgaudin:
> can you explain a little more about it?

I don't understand. Is there something specific about my previous comment that is not clear?

hksingh4545:
Is the file which is including during setup is different from current one?

kelson42:
@rgaudin any feedback?

Issue #160, opened 2023-01-19
Do not force the presence of PDF format for all books


benoit74:
A [few lines](https://github.com/openzim/gutenberg/blob/5be8c9b3fd6dfb61d69219a4fe1d703b5b8a9857/gutenbergtozim/rdf.py#L281-L288) are forcing the presence of the PDF format for all books.

This is causing broken links / weird buttons in all books which do not have a PDF format.

Discussion on this issue already started [here](https://github.com/openzim/gutenberg/issues/159) by mistake (my bad).

rgaudin:
I don't understand this ticket. If it's independent to #159, does that mean that we have PDF files in ZIM files always? Even when we don't request them?

benoit74:
It is relatively independent from #159 
It means that no matter the content of the RDF, the scraper consider there is always a PDF format available for all books.
This causes the behavior you saw while testing #163 that when you request PDF among other export formats, you get many books with broken PDF buttons because the format is supposed to be there according to the `bookformat` table.

rgaudin:
OK it's more clear: it fakes the availability of PDF for all books, even if there's no real PDF. Understood.

I don't see how it could affect #159 though but I believe we're not understanding each other on what #159 is about 😅

eshellman:
FWIW, pdf generation at PG is planned for sometime this year.

PR #163, opened 2023-01-19
Display buttons to the various book formats only for requested formats

### LINKED ISSUE #159: UI does not limit the buttons displayed to only requested formats
When we request at the CLI a list of formats (e.g epub + html) which is less than the list of formats of a given book (e.g. epub + html + pdf), only appropriate formats are exported in the ZIM (epub + html) but the UI shows a link to all existing formats (e.g. epub + html + pdf) the book supports (in the DB). There is hence broken links since we do not requested this additional format. Code seems to ignore this possibility. 

Here's a sample command to use:
```
python gutenberg2zim -z gut_fr_test.zim -b 18812 -l fr -f epub,html
```

Looks like the issue is present in many places: cover article (`gutenbergtozim/templates/cover_article.html`), HTML version of a book (`gutenbergtozim/templates/book_infobox.html`) and bookshelves (2 times in `gutenbergtozim/templates/js/tools.js`).

benoit74:
Fix #159 

Sample command used for tests (for some tests below, popularity of book 18813 has been tweaked in the extracted RDF in `rdf-files` to be bigger than the popularity of book 18812, so that sorting by popularity/title produces a different result):
```
python gutenberg2zim -z gut_fr_test.zim -b 18812,18813 -f epub,html
```

What has been checked while browsing the resulting ZIM:
- the home page displays two books with only two buttons for epub and html
- the cover page of a given book displays two books with only two buttons for epub and html
- the bookshelf of a given author displays one book with only two buttons for epub and html
- the HTML version of a given book displays only one button for epub
- sorting by popularity / title works as expected

### DIFF
--- a/gutenbergtozim/database.py
+++ b/gutenbergtozim/database.py
@@ -165,8 +165,8 @@ class Book(BaseModel):
             "cover_page": self.cover_page,
         }
 
-    def to_array(self):
-        fmts = self.formats()
+    def to_array(self, all_requested_formats):
+        fmts = self.requested_formats(all_requested_formats)
         return [
             self.title,
             # self.subtitle,
@@ -187,6 +187,11 @@ class Book(BaseModel):
         from gutenbergtozim.utils import main_formats_for
 
         return main_formats_for(self)
+
+    def requested_formats(self, all_requested_formats):
+        fmts = self.formats()
+        requested_formats = [fmt for fmt in fmts if fmt in all_requested_formats]
+        return requested_formats
 
 
 class BookFormat(BaseModel):
--- a/gutenbergtozim/export.py
+++ b/gutenbergtozim/export.py
@@ -41,7 +41,6 @@ from gutenbergtozim.utils import (
     get_langs_with_count,
     get_list_of_filtered_books,
     is_bad_cover,
-    main_formats_for,
     read_file,
     save_file,
     zip_epub,
@@ -323,7 +322,7 @@ def html_content_for(book, src_dir):
         raise
 
 
-def update_html_for_static(book, html_content, epub=False):
+def update_html_for_static(book, html_content, formats, epub=False):
     soup = BeautifulSoup(html_content, "lxml-html")
 
     # remove encoding as we're saving to UTF8 anyway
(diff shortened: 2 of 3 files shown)


rgaudin (changes requested):
Thank you but it's not working as expected. You are testing with books that don't have PDF so it's gonna be difficult 😉

```sh
python gutenberg2zim -z gut_fr_test_withoutpdf.zim -b 18812,18813,16816 -f epub,html
python gutenberg2zim -z gut_fr_test_withpdf.zim -b 18812,18813,16816 -f epub,html,pdf
```

It works as expected in the home page: it's not showing the PDF button anymore but on the content page (`Le roman de la rose - Tome I_cover.16816.html`) the button is still there… linking to a missing PDF.

benoit74:
Do you mean that the button is still there even when you request only `epub,html`? This is not what I get.
It is another known issue (#160) that the button for PDF is present on all books when you request `epub,html,pdf` even if the book does not have a PDF available on PG, but has nothing to do with the UI / ZIM creation stage, the issue is that the scraper forces the presence of a PDF format for all books at the parsing phase.

rgaudin:
> Do you mean that the button is still there even when you request only `epub,html`? This is not what I get.

Because you're not testing against the correct books. Try with the command I mentioned.

> It is another known issue (#160) that the button for PDF is present on all books when you request `epub,html,pdf` even if the book does not have a PDF available on PG, but has nothing to do with the UI / ZIM creation stage, the issue is that the scraper forces the presence of a PDF format for all books at the parsing phase.

Then what does this PR should be fixing? the ticket you're closing is named “UI does not limit the buttons displayed to only requested formats” and the PR itself is named “Display buttons to the various book formats only for requested formats”.
I am seeing a format-button (PDF) for a format I did not request. That's exactly what I understand should be fixed by this PR.
Now if you're saying PDF is there because of a different bug, why am I seeing the epub button as well in the book page when requesting only html?

benoit74:
I'm very sorry, but I do not see a format-button (PDF) when using the following command:
```
python gutenberg2zim -z gut_fr_test_withoutpdf.zim -b 18812,18813,16816 -f epub,html
```

Did you cleared the `dl-cache` folder? I forgot to mention that cover pages are generated in this folder and not regenerated if you request a different list of formats (yes, it sucks, but I'm not sure about how to fix this).

You are right that PDF button should be there only if requested, and this is what I observe in my tests.

What I mentioned is that when PDF is requested (and only then), the button will be there for all books even if there is no PDF available in reality, and this is the other issue. In other words, I mentioned that for now it is normal to have PDF buttons with broken links for books which do not have a PDF format in PG, this is the other issue. But this should happen only if PDF is requested.

rgaudin (approved):

Issue #145, opened 2022-01-04
Remove inline JavaScript to comply with some CSPs


Jaifroid:
This is the same issue in Gutenberg that has been reported for PhET here: https://github.com/openzim/phet/issues/134 . To summarize, Content Security Policies of browser extensions forbid running inline JavaScript. Our ZIMs should avoid having JS inline, and should put required blocks in separate js files. For example, in the landing page of Gutenberg ZIMs, we have:

`<body onload="init();   showBooks();   " class="pure-skin-gutenberg     home">`

This could easily be replaced with a much safer bit of code in the JS file (`tools.js`) where `init()` and `showBooks()` are defined. All that would be needed is something like:
```
document.querySelector('body').addEventListener('load', function() {
    init();
    showBooks();
};
```
For more info, see https://github.com/kiwix/kiwix-js/issues/789.

The effect of the current inline JS can be seen in our Chromium extension. Compare the left screenshot from the PWA version with the screenshot from the Chromium extension (both running in Service Worker mode, and both capable of running JS). The books are not shown in the extension because `init()` and `showBooks()` are blocked by the CSP.

![image](https://user-images.githubusercontent.com/4304337/148059704-204e7ed1-7436-4a97-bb73-624c5d7881f4.png)

eshellman:
The base.html template fills the onload attribute this way:
`onload="init();  {% if show_books %} showBooks(); {% else %} populateFilters(); {% endif %} {%if bookshelf_home %} showBookshelfSearchResults(''); {%endif%} {%if bookshelf is defined%} showBookshelf(`{{ bookshelf }} `); {%endif%}`
so the show_books, bookshelf_home and bookshelf attributes would need to be exposed to the javascript somehow - perhaps using `data-` attributes?

rgaudin:
Would you be able to submit a PR ?

What about:

https://github.com/openzim/gutenberg/blob/cb6a16c1e717d381a26c8593738ada45c98ca0fa/gutenbergtozim/templates/cover_article.html#L20

Also, it looks like there's an unused `<script>` in bookshelf.html.

Jaifroid:
@rgaudin That looks a bit like ReactJS to me, but it might be some other templating system... We realize it may not be possible to eliminate all inline scripting for ZIMs that rely heavily on custom UIs, especially those made with React. But if we can ensure a base level of accessibility for clients with restrictive CSPs, it would be really helpful.

rgaudin:
[...]

Issue #65, opened 2018-08-29
Add translation for Finnish


dattaz:
We want to have our interface for gutenberg project in Finnish.
File with strings to translate in json is here https://github.com/openzim/gutenberg/blob/master/gutenbergtozim/l10n.py and you should add you locale here : https://github.com/openzim/gutenberg/blob/master/gutenbergtozim/export.py#L40

kelson42:
We should interface this scraper with Translatewiki.

kelson42:
Basically a duplicate of #56

Issue #29, opened 2014-10-12
Some books have no pictures


EduAir:
In some books we see white squares which meens that the images are not loaded.

PR #173, opened 2023-02-22
Remove unused variables from tool.js

### LINKED ISSUE #158: Resolve CodeFactor issues on tools.js
CodeFactor is currently complaining about many issues in `templates/js/tools.js`

Most of them should be fixed. Usually the only one which are tolerated are the "complex method" ones (which would need a massive revamping).

See e.g. https://github.com/openzim/gutenberg/pull/157/checks?check_run_id=10676559369

theparthshukla:
fix #158 Removed unused variables, however left unused functions intact.

### DIFF
--- a/gutenbergtozim/templates/js/tools.js
+++ b/gutenbergtozim/templates/js/tools.js
@@ -1,7 +1,6 @@
 var sortMethod = "popularity";
 var booksUrl = "full_by_popularity.js";
 var bookshelves = "bookshelves.js";
-var inBooksLooadingLoop = false;
 var booksTable = null;
 var title_dict = null;
 var globalShelvesTable = null;
@@ -14,7 +13,7 @@ var persist_options = {
 };
 
 
-function queryParams(key) {
+/*function queryParams(key) {
   var qd = {};
   if (location.search) location.search.substr(1).split("&").forEach(function(item) {
     var s = item.split("="),
@@ -26,7 +25,7 @@ function queryParams(key) {
     return qd;
   else
     return qd[key];
-}
+}*/
 
 function getPersistedPage() {
   var pp = $("#page_record").val();
@@ -41,7 +40,7 @@ function getPersistedPage() {
   }
 }
 
-function getRequestedPage() {
+/*function getRequestedPage() {
   var qp = queryParams("page");
   try {
     return parseInt(qp) - 1;
@@ -52,7 +51,7 @@ function getRequestedPage() {
     }
     return 0;
   }
-}
+}*/
 
 function onTablePageChange(e, settings, table) {
   // record global ref to table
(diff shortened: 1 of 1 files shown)


benoit74:
LGTM from my mobile, I will run a more detailed review (including a test with a real ZIM) later to confirm this.

@rgaudin @kelson42 @theparthshukla : do you have experience of situations where old browsers would complain that JavaScript function are missing an argument?

Here an additional argument is passed by the caller but not used in the function, so we can remove it, but is it a recent convention or something which has almost always been supported in JavaScript ?

I'm a bit afraid of testing these JavaScript on recent modern browsers while we know that a significant portion of Kiwix/ZIM users are still on very old versions

kelson42:
This is more a question for our JS expert @Jaifroid

Jaifroid:
@benoit74 Gutenberg ZIMs can be read on old browsers so long as the user uses the native search to access book covers and the books themselves, and not the proprietary UI (which doesn't work in Kiwix JS versions if the browser doesn't support Service Workers, because only in Service Worker mode can the app run JS in the ZIM).

There is, however, a specific issue #145 with Chromium extensions (Chrome, Edge, etc.). These ban the use of inline JS, which means that the proprietary UI also doesn't work in the latest Chromium browsers **in the extension/add-on version**. The UI **does** work in the PWA version (pwa.kiwix.org).

Generally it's pretty trivial to change the use of inline JS by putting inline scripts into a separate .js files and simply loading them in the HTML.

kelson42:
What is the status here?

benoit74:
Never had time to review this in more details, and it looks like it will
not happen in the short term :-(

rgaudin:
Wouldn't fix the issue as there are additional unused variables in the file. Hint, codefactor listed 28, only 22 are fixed here. Please complete.

rgaudin (changes requested):
Wouldn't fix the issue as there are additional unused variables in the file. Hint, codefactor listed 28, only 22 are fixed here. Please complete.

kelson42:
@theparthshukla Any news?

theparthshukla:
@kelson42 whoops, the notification seems to have missed my mailbox. Will get to this later today

theparthshukla:
Line #65 ‘goToAuthor’ is defined but not used. However, it is actually used in “cover_article.html” on line #20. 

Line #392 'showBookshelf' is defined but never used. However, it is actually used in “base.html” on line #29.

Line #611 ‘init’ is defined but never used. However, it is actually used in “base.html” on line #29.

These were 3 issues of the 28 that were not taken care of as they were incorrectly identified.

Upon removal of the function “queryParams” as it was a function made redundant by the removal of the function “getRequestedPage”, (not identified in the initial issues) we get a total of 26 issues fixed.

rgaudin (approved):
Thank you! Works fine 👍

PR #221, opened 2024-03-05
Do not insert all RSYNC paths in database

### LINKED ISSUE #220: All RSYNC paths are inserted as Url in DB
When we pass the `--book` CLI argument, only few books will be processed.

However, in the `parse` phase the scraper logic still inserts all paths found in RSYNC result as potential URL. This leads to enormous waste of time, typically when debugging.

benoit74:
Fix #220 

Changes (all for the stage where we insert RSYNC result as Url into database):
- ignore all directory entries
- ignore all entries in an /old/ subfolder
- take into account the book selection which might have been passed
- display statistics after operation

### DIFF
--- a/src/gutenberg2zim/urls.py
+++ b/src/gutenberg2zim/urls.py
@@ -226,7 +226,7 @@ def build_html(files):
     return list(set(urls))
 
 
-def setup_urls(force):
+def setup_urls(force, books):
     file_with_url = TMP_FOLDER_PATH.joinpath(f"file_on_{UrlBuilder.SERVER_NAME}")
 
     if file_with_url.exists() and not force:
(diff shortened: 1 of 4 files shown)


src/gutenberg2zim/urls.py
```diff
@@ -261,10 +260,32 @@ def setup_urls(force):
...
+            if line.startswith("d"):
+                count_dir += 1
+                continue
+            # ignore all entries in an /old/ subfolder
+            if "/old/" in line:
+                count_old += 1
+                continue
+            # take into account the book selection which might have been passed
+            if books:
+                if not any(f"/{book}/" in line for book in books):
```
rgaudin, line 280: Do you have a sample of that rsync file? I remember deep nested trees and want to make sure that if we have books=[1] we don't include Book 111 because it has a path like `xxxxx/111/1/zzz`

rgaudin (approved):
LGTM ; please check my concern before merging

src/gutenberg2zim/urls.py
benoit74: Sure: https://tmp.kiwix.org/ci/test-gutenberg/file_on_aleph_pglaf_org
You are right regarding book IDs 1 to 9, this enhancement to make the scraper faster will not work, we will process too many URLs for nothing. But honestly, I do not care, in > 99.9% of the cases this PR will bring big enhancements. And it only means we should avoid using book IDs 1 to 9 for debugging / testing, or we will (as of today) process many unneeded URLs (it is only about which URL we add to the DB, the rest of the processing logic is unchanged, so it won't produce any final bug in the ZIM). And we will probably get rid of this rsync thing anyway.

rgaudin: OK ; then add a small comment explaining it's a known limitation

benoit74: Done

Issue #206, opened 2023-08-21
Gutenberg logs are all under the same `name`


benoit74:
In Gutenberg logs, only one logger `name` (`gutenberg2zim.constants`) is used making it pretty useless.

```
[gutenberg2zim.constants::2023-08-19 11:40:30,563] INFO:	Parsing file cache/epub/99/pg99.rdf for book id 99
[gutenberg2zim.constants::2023-08-19 11:40:31,442] INFO:	Parsing file cache/epub/9/pg9.rdf for book id 9
[gutenberg2zim.constants::2023-08-19 11:40:32,515] INFO:Add possible url to db
[gutenberg2zim.constants::2023-08-19 11:40:32,517] DEBUG:bash -c rsync -a --list-only rsync://aleph.pglaf.org/gutenberg/ > tmp/file_on_aleph_pglaf_org
```

We should not log the `name` anymore and instead log the filename with `%(filename)s` or module with `%(module)s`

rgaudin:
We've found that a single name is enough in most scrapers so we use the name to distinguish our logs from the other dependencies. Here it should use `gutenberg2zim` instead of the module name.
We could use different name base on file or module but it brings little value and make the logs very difficult to read because lines are not aligned (prefix size changes)

elfkuzco:
@benoit74 , I would like to implement this. Should I stick with keeping the module names or just use `gutenberg2zim` as @rgaudin suggested?

benoit74:
Just use one name, `gutenber2zim` as suggested by @rgaudin

And please adapt the code to create the logger with scraperlib `getLogger` function like we try to harmonize among our codebase. 

One good example of this approach is in offspot/demo:
- create the logger in constants.py (or something like that) with `getLogger`:  https://github.com/offspot/demo/blob/669163c8bc864a0120ac8b31f3c8c73c57acb55e/src/offspot_demo/constants.py#L53
- adjust the log level if debug/verbose mode is activated : https://github.com/offspot/demo/blob/669163c8bc864a0120ac8b31f3c8c73c57acb55e/src/offspot_demo/watcher.py#L76

Thank you!

benoit74:
Fixed by https://github.com/openzim/gutenberg/pull/223

PR #223, opened 2024-04-01
add flag to show verbose output and log output under one name

elfkuzco:
## Rationale
Normalize all log outputs to be under one name instead of module specific loggers and add `--debug` flag for more verbose output. This is intended to be a fix for #206 .

## Changes
- Initialize logger with default level of `INFO` instead of `DEBUG`
- Add `--debug` flag to dynamically set level to `DEBUG`
- When the `--debug` flag is set, update the handlers to have their levels set to `DEBUG` as a simple `logger.setLevel(logging.DEBUG)` only sets it for the logger and not for the handlers. This is akin to what the `getLogger` function from `zimscraperlib` does. After all, it is the handlers that handle output and not the `logger` object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,17 @@ as of 2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Add support for `--debug` flag to output debug logs
+
 ### Changed
 
 - Insert as few rsync URLs as possible in DB when a book selection is made (#220)
+
+### Fixed
+
+- Simplify the logger named (used `gutenberg2zim` instead of `gutenberg2zim.constants`) (#206)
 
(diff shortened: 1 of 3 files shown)


benoit74 (approved):
> When the --debug flag is set, update the handlers to have their levels set to DEBUG as a simple logger.setLevel(logging.DEBUG) only sets it for the logger and not for the handlers. This is akin to what the getLogger function from zimscraperlib does. After all, it is the handlers that handle output and not the logger object.

You are damn right! Code in offspot/demo I pointed to you is just not working as expected! Thank you!

I amended your commit to add a CHANGELOG entry. Please do not forget (try to at least, I tend to forget as well) next time.

PR #225, opened 2024-04-13
Replace usage of os.path and path.py with pathlib

### LINKED ISSUE #195: Remove usage of os.path and path.py, prefer pathlib
We use both `os.path`, `path.py` and `pathlib` in the codebase.

This is pretty confusing in many situation. We should use only one, and `pathlib` is the one to choose.

elfkuzco:
Fixes #195 

## Rationale

Use `pathlib.Path` for all path operations instead of a combination of `os.path` and `path.py`.

## Changes
- Replace all usages of `os.path` and `path.py` to `pathlib.Path`
- Add type annotations for some parts of the code
- Use f-strings to build urls instead of `os.path.join`

### DIFF
--- a/src/gutenberg2zim/download.py
+++ b/src/gutenberg2zim/download.py
@@ -1,17 +1,15 @@
-import os
-import pathlib
 import shutil
 import tempfile
 import zipfile
 from multiprocessing.dummy import Pool
+from pathlib import Path
 from pprint import pprint as pp
 
 import apsw
 import backoff
 from kiwixstorage import KiwixStorage
-from path import Path
 
-from gutenberg2zim.constants import TMP_FOLDER, logger
+from gutenberg2zim.constants import TMP_FOLDER_PATH, logger
 from gutenberg2zim.database import Book, BookFormat
 from gutenberg2zim.export import fname_for, get_list_of_filtered_books
 from gutenberg2zim.s3 import download_from_cache
@@ -36,24 +34,24 @@ IMAGE_BASE = "http://aleph.pglaf.org/cache/epub/"
 #         return False
 
 
-def handle_zipped_epub(zippath, book, dst_dir: pathlib.Path):
+def handle_zipped_epub(zippath: Path, book: Book, dst_dir: Path) -> bool:
     def clfn(fn):
-        return os.path.join(*os.path.split(fn)[1:])
+        return Path(fn).name
 
     def is_safe(fname):
-        fname = ensure_unicode(clfn(fname))
-        if Path(fname).basename() == fname:
+        name = ensure_unicode(clfn(fname))
+        if Path(fname).name == name:
             return True
-        return fname == os.path.join("images", Path(fname).splitpath()[-1])
+        return fname == f"images/{Path(fname).name}"
 
     zipped_files = []
     # create temp directory to extract to
-    tmpd = tempfile.mkdtemp(dir=TMP_FOLDER)
+    tmpd = tempfile.mkdtemp(dir=TMP_FOLDER_PATH)
     try:
         with zipfile.ZipFile(zippath, "r") as zf:
             # check that there is no insecure data (absolute names)
             if sum([1 for n in zf.namelist() if not is_safe(ensure_unicode(n))]):
-                Path(tmpd).rmtree_p()
+                shutil.rmtree(tmpd, ignore_errors=True)
                 return False
             # zipped_files = [clfn(fn) for fn in zf.namelist()]
             zipped_files = zf.namelist()
@@ -64,7 +62,7 @@ def handle_zipped_epub(zippath, book, dst_dir: pathlib.Path):
         # file is not a zip file when it should be.
         # don't process it anymore as we don't know what to do.
         # could this be due to an incorrect/incomplete download?
-        return
+        return False
 
     # is there multiple HTML files in ZIP ? (rare)
     mhtml = (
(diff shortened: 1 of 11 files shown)


src/gutenberg2zim/download.py
```diff
@@ -329,10 +332,9 @@ def download_cover(book, book_dir, s3_storage, optimizer_version):
         etag = get_etag_from_url(url)
         downloaded_from_cache = False
         cover = f"{book.id}_cover_image.jpg"
-        if (
-            book_dir.joinpath("optimized").joinpath(cover).exists()
-            or book_dir.joinpath("unoptimized").joinpath(cover).exists()
-        ):
+        if (book_dir / "optimized" / cover).exists() or (
```
benoit74, line 335: This is hard to read, please add back the outermost parentheses if possible, so that code is written more clearly

src/gutenberg2zim/entrypoint.py
```diff
@@ -218,21 +222,25 @@ def f(x):
     for zim_lang in zims:
         if do_zim:
             logger.info("BUILDING ZIM dynamically")
+            if one_lang_one_zim_folder:
+                output_folder = Path(one_lang_one_zim_folder).resolve()
+            else:
+                output_folder = Path(".").resolve()
             build_zimfile(
-                output_folder=Path(one_lang_one_zim_folder or ".").abspath(),
+                output_folder=output_folder,
```
benoit74, line 230: We prefer to avoid assigning variables which are used only once, this proved to be confusing for next maintainers. 

You should rewrite this `output_folder = Path(one_lang_one_zim_folder).resolve() if one_lang_one_zim_folder else Path(".").resolve()`

src/gutenberg2zim/export.py
```diff
@@ -94,8 +93,8 @@ def save_bs_output(soup, fpath, encoding=UTF8):
 jinja_env.filters["urlencode"] = urlencode
 
 
-def tmpl_path():
-    return os.path.join(Path(gutenberg2zim.__file__).parent, "templates")
+def tmpl_path() -> Path:
+    return (Path(gutenberg2zim.__file__).parent / "templates").resolve()
```
benoit74, line 97: Do we really need to `resolve` here? `gutenberg2zim.__file__` is already resolved

src/gutenberg2zim/download.py
```diff
@@ -73,25 +71,26 @@ def is_safe(fname):
     # move all extracted files to proper locations
     for zipped_file in zipped_files:
         # skip folders
-        if not Path(zipped_file).ext:
+        if not Path(zipped_file).resolve().is_file():
```
benoit74, line 74: How can it work? `zipped_file` is path inside the ZIP, isn't it?

src/gutenberg2zim/download.py
```diff
@@ -73,25 +71,26 @@ def is_safe(fname):
...
             continue
 
-        src = os.path.join(tmpd, zipped_file)
-        if os.path.exists(src):
-            fname = Path(zipped_file).basename()
+        src = (Path(tmpd) / zipped_file).resolve()
```
benoit74, line 77: Do we really need to `resolve` here? `tmpd` is already resolved

src/gutenberg2zim/rdf.py
```diff
@@ -297,9 +305,9 @@ def get_formatted_number(num):
     nums = [f"{i:0=5d}" for i in range(21000, 40000)]
     for num in nums:
         print(num)  # noqa: T201
-        curd = os.path.dirname(os.path.realpath(__file__))
-        rdf = os.path.join(curd, "..", "rdf-files", num, "pg" + num + ".rdf")
-        if os.path.isfile(rdf):
+        curd = Path(__file__).resolve().parent
```
benoit74, line 308: Do we really need to `resolve` here? `__file__` is already resolved

src/gutenberg2zim/urls.py
```diff
@@ -291,5 +290,5 @@ def setup_urls(force, books):
 
 
 if __name__ == "__main__":
-    book = Book.get(id=9)
+    book = Book.get(id=84)
```
benoit74, line 293: Do we really need to change this? Book id 9 is wrong?

benoit74 (changes requested):
Thank you, this looks very good, only minor comments to discuss (not sure I'm right on all of them ^^)

src/gutenberg2zim/download.py
elfkuzco: Okay

src/gutenberg2zim/entrypoint.py
elfkuzco: Okay. Will do that

src/gutenberg2zim/export.py
elfkuzco: Yeah, you are right. That is absolutely redundant

src/gutenberg2zim/download.py
elfkuzco: According to the comment, it should skip folders, hence the `is_file()` call. For the call to `resolve()`, I was experimenting with the list of items returned from a zip file and they are all relative to the zipped file. However, looking at it now, I think it should work without the `resolve()` call

src/gutenberg2zim/download.py
elfkuzco: While rewriting the code, I didn't know exactly which folders were passed as arguments and started resolving from the bottom up. But now that I know better, I certainly think I can do better by removing the redundant `resolve()` calls. Thanks for pointing that out.

src/gutenberg2zim/urls.py
elfkuzco: Oh, my bad. This is when I was testing the `urls.py` script after downloading book 84. Another oversight.

src/gutenberg2zim/download.py
elfkuzco: Tried to add the outermost parentheses but `black` keeps reformatting it.

elfkuzco:
@benoit74 , I have made the necessary fixes but I'm unable to run the full check anymore as my VM instance on GCP keeps running out of space (over 20Gb). However, I can verify that the previous commands run to full completion

benoit74:
> I have made the necessary fixes but I'm unable to run the full check anymore as my VM instance on GCP keeps running out of space (over 20Gb). However, I can verify that the previous commands run to full completion

That's enough for me, we will make a full run before releasing this anyway.

benoit74 (approved):
LGTM

Issue #227, opened 2024-09-03
Clarify GNU/Linux requirements on README.md


benoit74:
Two problems:
- the command works only on apt-based distro, so it is not for GNU/Linux
- some packages (like `python-dev`, but probably others as well) do not really seem to be required, especially if user only want to run the scraper and not modify it

PR #234, opened 2025-03-14
Update README and PyPI descriptions; add non-Debian support details

cking100:
This PR  fixes the issue #227 by  updating the README and PyPI description. The main focus is to enhance clarity, fix any inconsistencies because some packages are not required when u only use scraper and include information relevant to non-Debian-based distributions like Arch and Fedora. These changes help ensure that users on different Linux distributions have the necessary details for installation and usage.

README Updates:
Added information about compatibility with Arch, Fedora, and other non-Debian distributions.
also used 
i have used  "sudo apt update && sudo apt install -y python3-pip libxml2-dev libxslt-dev advancecomp jpegoptim pngquant p7zip-full gifsicle curl zip zim-tools" which makes sure all the packages are up to date and used python3 because the previous ones where outdated and did not support arch and fedora
Also changes the pypi-readme.rst for support of arch and fedora which made the overall readme more readable and concise

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ To view all the available options for `gutenberg2zim`, run:
 docker run ghcr.io/openzim/gutenberg:latest gutenberg2zim --help
 ```
 
+
 ### Arguments
 
 Customize the content download with the following options. For example, to download books in English or French with IDs 100 to 200 and only in PDF format:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -82,56 +155,6 @@ You can find the full arguments list below:
...
-### Setup the package
-
-First, clone this repository.
-
-```bash
-git clone git@github.com:kiwix/gutenberg.git
-cd gutenberg
-```
-
-If you do not already have it on your system, install `hatch` to build the software and manage virtual environments (you might be interested by our detailed [Developer Setup](https://github.com/openzim/_python-bootstrap/blob/main/docs/Developer-Setup.md) as well).
```
Murdock9803, line 120: `"you might be interested by"` should be changed to `"you might be interested in"`. Minor grammatical error 🙌

README.md
```diff
@@ -32,6 +32,79 @@ The `-v $(pwd)/output:/output` option mounts the `output` folder in your current
...
+
+Start a hatch shell: this will install software including dependencies in an isolated virtual environment.
+
+```bash
+hatch shell
+=======
+1. **Run the scraper with Docker**:
+
+```bash
+docker run -it --rm -v $(pwd)/output:/output ghcr.io/ope
```
Murdock9803, line 99: Shouldn't this command be `docker run -it --rm -v $(pwd)/output:/output ghcr.io/openzim/gutenberg:latest gutenberg2zim` ? I think it is incomplete.

README.md
```diff
@@ -32,6 +32,79 @@ The `-v $(pwd)/output:/output` option mounts the `output` folder in your current
...
+```
+
+The `-v $(pwd)/output:/output` option mounts the `output` folder in your current directory to the `/output` folder inside the container (which is the working directory). This ensures that the ZIM file is saved to your local machine.
```
Murdock9803, line 102: This whole part is present in the file at the start too. Can we remove this to reduce excess text ?

cking100:
sure ill do the necessary changes

README.md
benoit74: This is your own modification ...

benoit74:
How did you came up with this list of packages on the various distros ? Did you tested the scraper works will in these context ? This is mandatory part of this PR.

The goal of the issue is also to trim down unnecessary packages. Here it looks like you are more adding all packages to all contexts, quite the opposite direction of the move we want to see happen.

benoit74:
@cking100 this would be perfect ! I will merge this PR since it is already a significant improvement, and please move forward with another PR once you've made it clear which dependencies are required for running the scraper vs developing it vs which one are not needed at all. Thanks a lot

PR #239, opened 2025-03-24
Fix incorrect ZIM title when multiple languages are present

### LINKED ISSUE #196: Fix "PLR2004" magic number in `urls.py`
`noqa: PLR2004` magic number hint has been added by #191 in `urls.py`

What is this magic number? To be expressed more clearly

### LINKED ISSUE #229: Title of "mul" ZIM is wrong
See:
Image

... it's not "Books in English"

Lorrainnn:
This PR addresses two issues:

- Fix #229 – ZIM metadata now correctly detects multilingual book lists and sets the mul language code accordingly, rather than defaulting to the first book’s language.

- Fix #196 – Replaced magic number 10 with a named constant and clarified its purpose in the comment.

image

### DIFF
--- a/src/gutenberg2zim/l10n.py
+++ b/src/gutenberg2zim/l10n.py
@@ -534,221 +534,279 @@ l10n_strings = {
 
 
 metadata_translations = {
+    "mul": {
+        "title": "Project Gutenberg Library",
+        "description": "All books in all languages "
+        "from the first producer of free Ebooks",
+    },
     "afr": {
-        "description": "Die eerste vervaardiger van gratis e-boeke",
-        "title": "Projek Gutenberg-biblioteek - Boeke in Afrikaans ",
+        "title": "Projek Gutenberg-biblioteek",
+        "description": "Alle boeke in Afrikaans "
+        "van die eerste vervaardiger van gratis e-boeke",
     },
     "ara": {
-        "description": "أول منتج للكتب الإلكترونية المجانية",
-        "title": "مكتبة مشروع جوتنبرج - كتب باللغة العربية",
+        "title": "مكتبة مشروع جوتنبرج",
+        "description": "جميع الكتب باللغة العربية من أول منتج للكتب"
+        " الإلكترونية المجانية",
     },
     "arp": {
-        "description": "The first producer of Free Ebooks",
-        "title": "Project Gutenberg Library - Books in Arapaho",
+        "title": "Project Gutenberg Nookóó",
+        "description": "Nii'eihi' núhu' núunóbe3en Arapaho nih'iitíínoo, "
+        "nih'iitokúuti' free Ebooks",
     },
     "bod": {
-        "description": "The first producer of Free Ebooks",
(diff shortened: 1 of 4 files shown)


src/gutenberg2zim/zim.py
```diff
@@ -50,10 +50,15 @@ def build_zimfile(
 
     formats.sort()
 
-    title = title or metadata_translations.get(iso_languages[0], {}).get(
+    if len(set(iso_languages)) > 1:
```
benoit74, line 53: - Prefer to use `mul` rather than `mult` 
- No need to compute the set (deduplication is - or should be - done before that)
- Naming the variable `metadata_lang` is clearer
- One liner is easier to read : `metadata_lang` = "mul" if len(iso_languages) > 1 else iso_languages[0]`

src/gutenberg2zim/l10n.py
```diff
@@ -534,6 +534,10 @@
 
 
 metadata_translations = {
```
benoit74, line 536: All these values have to be fixed to match [the spec](https://wiki.openzim.org/wiki/Metadata):
- title must not be longer than 30 graphemes
- description must not be longer than 80 graphemes

I propose to base the work on these English values (which have to be translated in all languages, using AI for this task is exceptionally acceptable):
- title: `Project Gutenberg Library`
- description: `All books in English from the first producer of free Ebooks`

`mul` is obviously an exception and we should use:
- title: `Project Gutenberg Library`
- description: `All books in all languages from the first producer of free Ebooks`

Default (when `iso_language` is missing from the `metadata_translations` dictionary could be:
- title: `Project Gutenberg Library`
- description: `All books in "${iso_language[0]}" language from the first producer of free Ebooks`

benoit74 (changes requested):
Thank you !

Please see inline comments.

More general changes needed as well:
- please add an entry in `CHANGELOG.md`
- please fix your first comment so that PR is properly linked to fixed issues: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

benoit74:
And please ensure that CI passes. You can invoke it locally with `invoke lintall`, `invoke checkall` and `invoke test`

Lorrainnn:
All updated. I manually double-checked all translations in title and description. Thx for suggestions, they are really comprehensive. 
Btw is there something I could do to better solve the CodeFactor Warning of "Complex Method" in line 120 in zim.py.
image

image

benoit74 (changes requested):
LGTM, thank you very much for this very efficient PR, it is very rare to achieve completion within only one review cycle, especially for a first-time contributor.

Now that review is complete, please squash all your commits into one so that we do not keep history of implementation details but only history of intended changes. I will merge it right after.

Lorrainnn:
Finished quashed the commits and updated the commit message as requested. Thanks again for the thorough review and encouragement! (ㅅ´ ˘ `)

PR #240, opened 2025-03-26
Replace print and pp calls with logger

### LINKED ISSUE #192: Logs are sometimes done with `print` or `pp`
Logs are sometimes done with `print` or `pp` which means they escape the logging mechanism allowing to filter messages.

For now #191 marks them with `# noqa: T201" and `# noqa: T203` but we have to decide what to do about those logs (especially what is the appropriate level, either debug, info, error, ...)

Lorrainnn:
- Fix #192 

pp being replaced with pformat:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ as of 2.0.0.
 - Replace usage of os.path and path.py with pathlib.Path (#195)
 - Finalize ZIM metadata title translations and multilingual detection (#229)
 - Replaced magic number with named constant and clarified comment regarding book ID URL rules (#196)
+- Replace print and pp calls with logger (#192)
 
 ### Fixed
 
--- a/src/gutenberg2zim/download.py
+++ b/src/gutenberg2zim/download.py
@@ -3,7 +3,7 @@ import tempfile
 import zipfile
 from multiprocessing.dummy import Pool
 from pathlib import Path
-from pprint import pprint as pp
+from pprint import pformat
 
 import apsw
 import backoff
@@ -91,11 +91,8 @@ def handle_zipped_epub(zippath: Path, book: Book, dst_dir: Path) -> bool:
             dst = dst.resolve()
             try:
                 src.rename(dst)
-            except Exception as e:
-                import traceback
-
-                print(e)  # noqa: T201
-                print("".join(traceback.format_exc()))  # noqa: T201
+            except Exception:
+                logger.error(f"Failed to move extracted file: {src} -> {dst}")
                 raise
 
     # delete temp directory and zipfile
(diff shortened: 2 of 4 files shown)


Lorrainnn:
I've also been digging into the root cause of the overall slowness—really appreciated the discussions in issue #97, they’re quite insightful lol!

Right now, I’m focusing on the parsing stage, particularly the setup_urls(force, books) step where URLs are appended to the database from the rsync result. This part takes noticeably longer than expected.

I've tested a revised approach using batch inserts and general update on efficiency, turns out the parsing time reduces 60%.

same command: python3 -m gutenberg2zim --books 12000-13000 --parse -y<br>
before
imageafter<br>
image

Should I open a new issue to talk about details of code for this or simply revise in this PR? :) Thank you so much!

Lorrainnn:
These are just my thoughts. Idk if it's appropriate to post here but you could tell it's actually very straightforward lmao.
```python
    count_dir = count_old = count_added = count_processed = 0
    #batch
    BATCH_SIZE = 1000
    batch = []

    #test
    start = time.time()
    book_patterns = [f"/{book}/" for book in books] if books else []
    with open(file_with_url, errors="replace") as src:
        
        for line in src:
            count_processed += 1
            # show progress in debug mode, we expect about 5.4M lines as of early 2024
            if count_processed and count_processed % 100000 == 0:
                logger.debug(f"\t{count_processed} rsync results processed")

       
            # ignore all directory entries
            if line.startswith("d"):
                count_dir += 1
                continue
            # ignore all entries in an /old/ subfolder
            if "/old/" in line:
                count_old += 1
                continue
            # take into account the book selection which might have been passed ;
            # this not does completely filter-out useless urls for books IDs 1 to 9
            # but still makes the scraper way faster for all other selections
            if book_patterns and not any(p in line for p in book_patterns):
                continue
            # strip rsync file to only contain relative path
            url = line[start_rel_path_idx:].strip()
            batch.append(Url(url=url))  # type: ignore
            count_added += 1

            if len(batch) >= BATCH_SIZE:
                Url.bulk_create(batch)
                batch.clear()
    #last batch
    if batch:
        Url.bulk_create(batch)
    logger.info(
        f"\tDB is ready, {count_added} URLs have been added ({count_dir} dirs ignored, "
        f"{count_old} old stuff ignored, {count_processed} lines processed)"
    )
    end = time.time()
    print(f"Elapsed time: {end - start:.2f} seconds")

benoit74:
Regarding your enhancement around parsing, it would have been better to open a dedicated issue, but no big deal. I've opened https://github.com/openzim/gutenberg/issues/241, and I will reply you there.

benoit74 (changes requested):
One small change required + points below

Other than that:
- please fix your first comment so that PR is properly linked to Issue (see https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue)
- please add an entry in CHANGELOG.md
- please squash all your commits into one commit (they look more like developments steps than real changes needed)

Lorrainnn:
Finished.

benoit74 (approved):
LGTM, thank you

PR #242, opened 2025-03-28
Simplify retrieval of book full-text, get rid of rsync, prefer epub3 to epub

### LINKED ISSUE #97: Simplify Gutenberg scraping (no more rsync, no more fallback URLs / filenames)

### LINKED ISSUE #160: Do not force the presence of PDF format for all books

### LINKED ISSUE #219: regression: Missing HTML content
In [gutenberg_en_all_2024-02](https://library.kiwix.org/viewer#gutenberg_en_all_2024-02/Home.html), out of the 10 books listed (all declaring offering an HTML version), **only three** do have an HTML version.

Screenshot 2024-03-05 at 10 06 08

Either HTML version it is not present in the ZIM or the link is incorrect (it's same link in listing and in preview page). This is not limited to those 10 entries but it makes this 75GB ZIM look like garbage.

Initially reported by Offspot user.

### LINKED ISSUE #235: Prefer EPUB 3 to EPUB
Project Gutenberg is now (probably since few years) proposing EPUB 3 in addition to EPUB (EPUB 2 in fact).

Should we prefer this "new" format?

EPUB3:
- standardized in October 2011
- supported by epub.js (the viewer used in kolibri)

ChatGPT input:

```
In 2025, using EPUB 3 instead of EPUB 2 (often called "standard" EPUB) has far fewer compatibility issues than in the past, but there are still some considerations regarding device coverage:
Disadvantages of EPUB 3 in Terms of Device Compatibility

    Older E-Readers May Not Fully Support EPUB 3
        Some legacy e-readers, particularly older Kindle models and older Nook devices, may struggle with EPUB 3 features.
        Certain EPUB 3 elements, like embedded multimedia (audio/video) or JavaScript-based interactivity, may not work properly.

    Amazon Kindle Conversion Issues
        Amazon’s Kindle ecosystem does not natively support EPUB (it uses AZW/KFX format). While Amazon now allows EPUB uploads, it converts them to its own format.
        EPUB 3 files may not convert perfectly through Kindle’s system, especially if they use advanced features.

    Potential Rendering Variability
        EPUB 3 offers more advanced typographic and layout options (e.g., fixed layouts, better CSS support), but some e-readers may render these inconsistently.
        Devices that rely on basic EPUB 2 support may display EPUB 3 content in a degraded manner (e.g., losing styling or interactivity).

Reasons to Still Choose EPUB 3

    EPUB 3 is the current industry standard and is widely supported by major platforms like Apple Books, Kobo, Google Play Books, and modern e-readers.
    It provides better accessibility, improved typography, and native support for modern web technologies like HTML5 and CSS3.
    It ensures future-proofing, as EPUB 2 is no longer being updated.

Conclusion

If your target audience includes older e-readers (especially non-updated e-ink devices), sticking to EPUB 2 might ensure broader compatibility. However, for most modern devices and platforms, EPUB 3 is the better choice in 2025.

Would you like advice on optimizing EPUB 3 for maximum compatibility?
```

benoit74:
Fix #97 
Fix #219
Fix #235
Fix #160

Changes:
- get rid of initial rsync
- get rid of `Url` DB table
- get rid of `BookFormat` DB table
- add `unsupported_formats` column in `Book` DB table, containing a CSV of formats known to not be present for a given book (e.g. `pdf,epub`)
- add and use @eshellman logic from https://github.com/gutenbergtools/libgutenberg/blob/master/pg_archive_urls.py (patched a bit to please ruff, ~~I will~~ I've open a PR in this repo: https://github.com/gutenbergtools/libgutenberg/pull/47)
- adapt the download logic (and a bit of the rest) accordingly


rgaudin (approved):
LGTM ; not tested.

I saw a mention of multiple passes ; do we still do that? How so?

benoit74:
> not tested.

Of course, will be done before publishing ZIMs again anyway

> I saw a mention of multiple passes ; do we still do that? How so?

Where ? We don't AFAIK ^^

src/gutenberg2zim/download.py
```diff
@@ -124,197 +139,117 @@ def download_book(
...
+
+    unoptimized_dir.mkdir(parents=True, exist_ok=True)
+    optimized_dir.mkdir(parents=True, exist_ok=True)
+
+    unsupported_formats = []
     for book_format in formats:
-        unoptimized_fpath = unoptimized_dir / fname_for(book, book_format)
+        logger.debug(f"Processing {book_format}")
+
+        # if we already know (e.g. due to a former pass) that this book format is not
```
rgaudin, line 152: here

benoit74: OK, got it.

We don't do it on the Zimfarm. But scraper is mostly capable to support it (e.g. if you run the tool twice with different settings while keep temporary files) so I ensured I did not introduced any regression on this "functionality".

PR #243, opened 2025-04-04
Upgrade scraplib to 5.x

### LINKED ISSUE #188: Upgrade python-scraperlib to 3.x, including CLI support for description / long_description flags
See https://github.com/openzim/overview/issues/21

### LINKED ISSUE #199: Rename Book DB table fields which are shadowing Python builtin
#191 added `#noqa: A003` in Book class used to represent Book DB table because `id` and `license` are shadowing Python builtin. 

Those fields should be renamed (at least license, id is very meaningful and maybe ok).

Lorrainnn:
Fix #188
Fix #199

- Update python-scraperlib to 5.1.1
- Update general dependency
- Add CLI support for description / long_description flags
- Rename Book DB table fields: id->book_id/license->license_id

Since updating to the new versions of Black and Ruff, some differences in how blank spaces are handled in the templates.


Lorrainnn:
Not yet update Zimfarm configuration. If this logic looks good to you, I will try to add new CLI paramters and squash the commits history later.

New zim logic has been fully tested, including `compute_description()`:
image

src/gutenberg2zim/export.py
```diff
@@ -200,7 +200,7 @@ def export_all_books(
         )
 
     # export illustation
-    export_illustration()
+    # export_illustration()
```
benoit74, line 203: Looks like a debug leftover ^^

src/gutenberg2zim/database.py
```diff
@@ -119,30 +119,31 @@ class Book(BaseModel):
     class Meta:
         database = db
 
-    id = IntegerField(primary_key=True)  # noqa: A003
+    book_id = IntegerField(primary_key=True)
     title = CharField(max_length=500)
     subtitle = CharField(max_length=500, null=True)
     author = ForeignKeyField(Author, related_name="books")
-    license = ForeignKeyField(License, related_name="books")  # noqa: A003
+    license_id = ForeignKeyField(License, related_name="books")
```
benoit74, line 126: I'm sorry, I did not realized this is a `ForeignKeyField`, and hence not only an id but the whole object.

I think that even if a bit ugly, `book_license` is probably easier to read / understand than `license_id`. WDYT?

src/gutenberg2zim/entrypoint.py
```diff
@@ -222,6 +228,12 @@ def f(x):
     for zim_lang in zims:
         if do_zim:
             logger.info("BUILDING ZIM dynamically")
+            description, long_description = compute_descriptions(
```
benoit74, line 231: You should probably do it way earlier in this method for two reasons:
- no need to compute this for every ZIM, value seems "static" for all ZIMs
- should this fails (it checks for proper length of description / long description) then the sooner the better

FYI, I've opened https://github.com/openzim/gutenberg/issues/244 which is kinda related

src/gutenberg2zim/export.py
```diff
@@ -505,16 +507,21 @@ def replacablement_link(book, url):
...
-    cover_img = f"{book.id}_cover_image.jpg"
+    cover_img = f"{book.book_id}_cover_image.jpg"
     cover_img = cover_img if (optimized_files_dir / cover_img).exists() else None
+
     translate_author = (
         f' data-l10n-id="author-{book.author.name().lower()}"'
         if book.author.name() in ["Anonymous", "Various"]
         else ""
     )
+    from gutenberg2zim.database import License
```
benoit74, line 518: Why do you need to do that now  ? Doesn't `book.license_id` (or `book.book_license` given my above comment) work?

CHANGELOG.md
```diff
@@ -19,6 +20,8 @@ as of 2.0.0.
 - Finalize ZIM metadata title translations and multilingual detection (#229)
 - Replaced magic number with named constant and clarified comment regarding book ID URL rules (#196)
 - Replace print and pp calls with logger (#192)
+- Update python-scraperlib to 5.1.1 and dependencies (#188)
```
benoit74, line 23: Please mention upgrade to Python 3.13 as well

benoit74 (changes requested):
Thank you, very good work, looking at everything you had to do I realize it was definitely not a small thing, especially for someone outside the core team developing the zimscraperlib. Hope you still enjoyed it a bit ^^

Small remarks / changes needed, but everything moved in the right direction

src/gutenberg2zim/database.py
Lorrainnn: book_license indeed better

src/gutenberg2zim/export.py
Lorrainnn: Agree.

src/gutenberg2zim/entrypoint.py
Lorrainnn: I see I see. Put it earlier.

benoit74 (changes requested):
Few more points:
- can't we remove the `export_illustration` method since we don't use it anymore ?
- please don't add back `jquery-1.11.1.min.js` ; this file should be in the gitignore in fact, since it is meant to be downloaded from the internet since https://github.com/openzim/gutenberg/pull/238 (I wasn't cautious enough in this review) ; we will download all those dependencies from the Internet soon.

Lorrainnn:
Finished.

benoit74 (approved):
LGTM

PR #247, opened 2025-04-09
Rebuild on DB for multiple languages cases and add request timout 

### LINKED ISSUE #197: Fix `requests` call without timeout in `utils.py`
`# noqa: S113` hint has been added by #191 in `utils.py`

This hint detected that we are using `requests` without specifying a timeout.

We should add one. To be decided: what is the proper timeout. Probably to be fixed together with #142.

### LINKED ISSUE #218: Gutenberg books can have multiple languages
See https://github.com/openzim/zim-requests/issues/841

Most probable explanation is that the scraper does not handle well the situation where a book has multiple associated languages.

Lorrainnn:
Fix #218 
Fix #197 

->solve gutenberg2zim -l ko. now able to handle multi lang in separate DB as booklanguage
image

### DIFF
--- a/src/gutenberg2zim/export.py
+++ b/src/gutenberg2zim/export.py
@@ -16,7 +16,7 @@ from six import text_type
 
 import gutenberg2zim
 from gutenberg2zim.constants import TMP_FOLDER_PATH, logger
-from gutenberg2zim.database import Author, Book
+from gutenberg2zim.database import Author, Book, BookLanguage
 from gutenberg2zim.iso639 import language_name
 from gutenberg2zim.l10n import l10n_strings
 from gutenberg2zim.s3 import upload_to_cache
@@ -96,7 +96,9 @@ def tmpl_path() -> Path:
 
 
 def get_list_of_all_languages():
-    return list({b.language for b in Book.select(Book.language)})
+    return list(
+        {bl.language_code for bl in BookLanguage.select(BookLanguage.language_code)}
+    )
 
 
 def export_skeleton(
@@ -177,6 +179,8 @@ def export_all_books(
     books = get_list_of_filtered_books(
         languages=languages, formats=formats, only_books=only_books
     )
+
+    logger.info(f"Found {books.count()} books for export")
 
     if not len(get_langs_with_count(books=books)):
         critical_error(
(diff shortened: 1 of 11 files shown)


Lorrainnn:
Hi! :)
1. Maybe we should still include /jquery/jquery-1.11.1.min.js right now since the scraper won't work as expected without the file. or to let this to be downloaded remotely ASAP.
2.  Apologies for the noise — I completely forgot to set the PR as a draft and ended up pushing a bit too often :(
By the way, should I worry about the Codecov bot report, or is that something to review later?

src/gutenberg2zim/export.py
```diff
@@ -877,6 +887,9 @@ def write_book_presentation_article(
     formats,
 ):
     article_name = article_name_for(book=book, cover=True)
+    if article_name in Global.added_items:
```
benoit74, line 890: Could you please explain why we need this duplicate check ? Maybe an example where this happen is enough. I don't get the rationale behind this, and would like to avoid "mostly silent errors" in the future.

src/gutenberg2zim/utils.py
```diff
@@ -41,9 +41,10 @@ def fname_for(book: Book, book_format: str) -> str:
 
 def get_etag_from_url(url: str) -> str | None:
     try:
-        response_headers = requests.head(  # noqa: S113
-            url=url, allow_redirects=True
+        response_headers = requests.head(
+            url=url, allow_redirects=True, timeout=10
```
benoit74, line 45: - please store the `10` in a `DEFAULT_HTTP_TIMEOUT` constant (in this `constants.py` file) 
- please use this for all `requests` call (we have some `get` with a 20 secs timeout ; I think replacing them with 10 is OK)

benoit74 (changes requested):
Very good, thank you! Overall this work great.

Regarding the jquery file, this is indeed mandatory but it is automatically downloaded when you `hatch build` (for instance). Your "problem" is that you've checked-out and setup the repo (with `hatch shell`) before the change removing the jquery file, and now that it's gone you probably did not ran `hatch build` again. That being said, I think there is still a packaging issue.

Regarding codecov, never mind, this is a WIP (not really started in fact tbh) on this scraper.

Regarding `update_book_languages`, I would either:
- remove this method, put code back inline and not mind about complex method warning so that code is more consistent
- or (my preference) isolate all steps of `save_rdf_in_database` into dedicated methods (but these could stay in `rdf.py` since they are not meant to be reused somewhere else)

Note that I've opened https://github.com/openzim/gutenberg/issues/249 to follow-up, but I do not have the answer to that question and it is OK to merge without it.

src/gutenberg2zim/export.py
Lorrainnn: You are right lol. This logic was used in my former testing to debug on when Id 5739 generate duplicate copies.

Lorrainnn:
Finish

benoit74 (approved):
LGTM

PR #252, opened 2025-04-17
Add RTL support and language filtering in UI based on user CLIs

### LINKED ISSUE #248: Implement real support for rtl languages
The UI has been translated to Arabic, but RTL support is not implemented, making the UI a bit clunky.

We should set proper `direction` style on `body` when the UI language is changed.

And some CSS must be adapted so that text properly float to the left/right.

### LINKED ISSUE #249: Only include selected languages in language combobox
With https://github.com/openzim/gutenberg/pull/247, we correctly implemented support for books with multiple languages.

It however has the consequence that we now have some books attached to two languages, causing the UI to display the language selection drop-down even for single language ZIM.

For instance in the following `en` ZIM, I still have the drop-down with `Korean` or `English` because book 5739 has both languages.

Image

Is it OK to keep the dropdown in the UI or should the dropdown show only languages which have been selected when building the ZIM (i.e. only `English` for an `en` ZIM?

Lorrainnn:
Fix #248 
Fix #249 

**TEST:**
command: python3 -m gutenberg2zim -l ko,it --books 5739,1000 -y 
5739: english, korean
1000: Italian
image

**RTL support:**
image
image

(Definitely RTL could be better with better general UI. lmao

**Question:** 
Copyrighted does not have locales implementation. would this become a problem?
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ as of 2.0.0.
 - Add support for `-L` long_description flags
 - Add request timeout for util.py (#197)
 - Add Booklanguage DB to support multi-languages books (#218)
+- Add RTL support to UI (#248)
+- Add language filter to combobox for requested languages (#249)
 
 ### Changed
 
@@ -28,9 +30,11 @@ as of 2.0.0.
 - Update multi-resolution favicons (#165)
 
 
+
 ### Fixed
 
 - Simplify the logger named (used `gutenberg2zim` instead of `gutenberg2zim.constants`) (#206)
+
 
 ## [2.1.1] - 2024-01-17
 
--- a/src/gutenberg2zim/export.py
+++ b/src/gutenberg2zim/export.py
@@ -191,6 +191,7 @@ def export_all_books(
     # export to JSON helpers
     export_to_json_helpers(
         books=books,
+        languages=languages,
         formats=formats,
         project_id=project_id,
         add_bookshelves=add_bookshelves,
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ as of 2.0.0.
 - Add support for `-L` long_description flags
 - Add request timeout for util.py (#197)
 - Add Booklanguage DB to support multi-languages books (#218)
+- Add language filter to combobox for available ZIM languages (#249)
```
benoit74, line 17: WDYT of `for available languages` => `for requested languages` ?

benoit74 (changes requested):
For all CSS changes, I'm pretty sure you do need the extra rules when direction is RTL but simply use appropriate CSS property in existing rules. For instance, you can replace `text-align: left` by `text-align: start` and the browser will take care of choosing left or right depending on document direction. Same for margins, you can use `margin-inline-end` or `margin-block-end` or `margin-inline-start` or `margin-block-start` together with `writing-mode`. See e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end. Using these properties means less maintenance on our end since we will have less CSS rules (and risk of inconsistencies between ltr and rtl, since I'm pretty sure we will forget to check rtl ^^). For sure support in browsers is more recent, but this is OK for us. Do not hesitate to speak up if this does not work.

Also see inline comment regarding the changelog.

Regarding the overall fact that design sucks a bit ... you know this well ^^

Regarding the fact that copyrighted property is not internationalized, I've opened https://github.com/openzim/gutenberg/issues/253

Lorrainnn:
> See e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end

Incredible features. I used to manually write separate RTL/LTR rules and indeed always wondering why this is soo time-consuming and prone to tiny mistakes lmao

Lorrainnn:
Finished.

image

benoit74 (approved):
LGTM, thank you!

Issue #290, opened 2025-07-09
Definition for current book Interface


Lorrainnn:
1. what's cover_page for? 
2. If bookshelf feature is no longer being maintined, should we delete this (all related bookshelf codes)
<img width="929" height="591" alt="Image" src="https://github.com/user-attachments/assets/61b07e2c-8c5d-4fbf-b389-5b72b54d5da9" />

benoit74:
Btw, it probably means that current UI is not properly handling the case where a book has no cover image (quite rare, but it does happen, at least it did happen in the past).

benoit74:
@VikramAditya33 can you confirm that `cover_page` is in fact used as a boolean indicating the book has a cover image? If yes, please rename it to a more meaningful name + change type to `bool`

Issue #276, opened 2025-07-06
Refine star rating logic


Lorrainnn:
The current popularity or star rating logic assigns stars based on download count thresholds and relative ranking, ensuring a forced distribution from 5 stars down to 1 star. However, this leads to misleading results when all books have similarly high download counts (e.g. all above 10,000). In such cases, some highly popular books still receive only 1–2 stars, simply due to their relative position.

logic refers to 208-232 in `export.py`

Question:
Should we keep this forced-distribution logic, or consider an alternative approach that gives more popularity ratings when all items are objectively popular?

Options could include:
Using absolute download thresholds (e.g. ≥10k = 5 stars);
...

benoit74:
> However, this leads to misleading results when all books have similarly high download counts (e.g. all above 10,000). 

If distribution is relative and all books have similarly high download counts, then the star mechanism allow to differentiate them. I don't get what the problem is.

If we move to absolute thresholds, wouldn't we have many books all with 5 stars? Which would not really help reader to get a good grasp of what is popular.

I think a concrete example where current system is not appropriate would definitely help, 

I also don't get if this relative ranking is done at scrape time (i.e. based on the whole corpus of books which have been downloaded) or live everytime users have selected some books through filtering / searching.

Lorrainnn:
Thanks for the feedback! I see your point. I’ll stick with the current logic for now(The new PR is based on previous logic), but happy to revisit it later if we run into more cases where it feels off.

PR #293, opened 2025-07-10
UI revamp architecture fix

### LINKED ISSUE #282: Define UI architecture to efficiently access required datasets
Currently, with https://github.com/openzim/gutenberg/pull/274 merged, the UI downloads the `gutenberg.db` file from the ZIM and uses it locally. 

While this is quite convenient, I'm not convinced this is going to work: on real selections (e.g. `en` which has tens of thousands books if I remember well), but also on the `mul` ZIM which contains all gutenberg books, the DB is going to be quite huge ; this means the reader will have to download this database and store it in memory ; ZIMs are expected to work well on small devices like phones with limited amount of memory available ; I'm concerned such a huge DB in memory might crash the reader on many occasions.

If the ZIM is read from a kiwix-serve (like library.kiwix.org), then this will also have a significant impact on first UI loading, since downloading such a huge DB is definitely going to take time on a variety of connections.

We should made an attempt to build the whole DB (with all gutenberg books, since this is the worst case scenario) and use it inside a ZIM. No need to have all books downloaded / rendered / saved inside the ZIM, goal is only to assess the size of the DB and try to use this ZIM in a variety of devices.

@Lorrainnn can you please build such a ZIM and share it? I will help with testing on a variety of readers.

This is probably the most urgent issue to tackle since should we have to change this, it is going to impact significantly the UI architecture.

### LINKED ISSUE #283: Favorites are not persistent
While there is a tentative implementation of favorites in the new UI, I find current result very deceptive because favorites are not saved / persisted at all. Books which have been marked as favorite is stored only in reader/browser memory.

Saving such favorites (or any user data in fact) is a daunting topic because we do not have such functionality really implemented in readers. We cannot modify the ZIM, so we need to store this data "somewhere else". We have some scrapers which experimented with local storage, but this is a complex topic.

I consider that this feature is definitely not the most important one so I suggest that for now we remove the favorites functionality. And we go back to this topic once the new UI is ok for all core behaviors (searching, filtering, ...).

### LINKED ISSUE #291: Book card layout breaks with long title, author, or rating
Long book titles or author names overflow or misalign in the book card UI.

Image

Lorrainnn:
Fix #291 
Fix #283 <- delete Favorite Feature 
Fix #282  <- change as discussed SQLite + Json per book


Lorrainnn:
image

Lorrainnn:
image

Lorrainnn:
Question: Now the homepage.db is generated alongside  gutenburg.db? Do we want to delete it afterwards. Or simply directly export this db from memory with `Global.add_item_for` (like how we export json if this is approachable

Some db details/logics could be simplified/deleted under #290.

zimui/src/views/BookDetail.vue
```diff
@@ -99,57 +97,30 @@ const decreaseFontSize = () => {
...
+  } catch (err) {
     book.value = {
       id: -1,
       title: 'Not Found',
       author: '',
       rating: 0,
       description: ''
     }
-    console.warn('Not found book_id:', bookId)
+    console.warn('Book not found:', bookId)
```
benoit74, line 122: We will need more than a console warning, something must be displayed on the page to explain the situation to the user. Can be done in a distinct issue.

zimui/src/views/HomePage.vue
```diff
@@ -69,35 +78,19 @@ onMounted(async () => {
...
   })
 
-  const dbUrl = new URL('gutenberg.db', document.baseURI).href
+  const dbUrl = new URL('homepage.db', document.baseURI).href
   const res = await fetch(dbUrl)
   const buffer = await res.arrayBuffer()
   const db = new SQL.Database(new Uint8Array(buffer))
 
   const loadNextBooks = () => {
-    if (isLoading.value) return
```
benoit74, line 78: Why are you not using this `isLoading` property? It looks like it was used to avoid concurrent data loading. I don't see what has changed on this matter.

benoit74 (commented):
Some general remarks:

- No need to delete `homepage.db`
- All generated files (homepage DB and JSON files) must be generated in the tmp folder, where scraper expects to have plenty of space available
- it looks like you've implemented something for https://github.com/openzim/gutenberg/issues/276 but you are not mentioning it at all ... I need details to make an informed decision on this point
- we've agreed in https://github.com/openzim/gutenberg/issues/282 and in live discussions that we need to better understand how homepage.db is going to be smaller than gutenberg.db 115Mb file ; please make such a measurement (or even better, a test ZIM like you did previously)

I've also added few inline comments.

zimui/src/views/BookDetail.vue
Lorrainnn: Got it. I’ll open a follow-up issue to display a user-facing message for missing books.

Lorrainnn:
Finish change.
I ll look into details in #290 later.

Lorrainnn:
I also add some basic logic about search (only a few lines about fuzzy search)

One issue I encountered is the following line:
`// eslint-disable-next-line @typescript-eslint/no-explicit-any
  let db: any = null`
I haven’t found a reliable way to import the correct Database type. Once we finalize the architecture, we can revisit whether this needs a proper solution or can be left as-is.

PR #300, opened 2025-09-30
Remove any optimization logic + use of S3 cache

benoit74:
Fix #193 
Fix #259
Fix #263
Fix #288 
Partial solution to #297 

Changes:
- do not optimize Gutenberg assets anymore since they are already optimized upstream
- do not use an S3 optimization cache anymore since:
  - we do not have optimized files to store anymore
  - original files are too often updated on the upstream, cache would barely never be hit + storing again something available upstream provides little added value

PR #301, opened 2025-09-30
Change mirror and add CLI flag to set mirror URL

benoit74:
`aleph.pglaf.org` is pretty slow ATM and we have a volunteer which setup a new mirror at `gutenberg.mirror.driftle.ss`.

This mirror is way faster. It probably suffer from few minutes / hours offset in terms of sync, but we should probably not care.

This change is obviously not definitive, even if it might stay for quite a long time, at least it unblocks current situation where scrapes take long in part due to this server being very slow.

This PR also adds a `--mirror-url` flag to allow to customize mirror used without having to release scraper again, and ensures we have a homogeneous use of same mirror throughout the whole processing.

To be merged after https://github.com/openzim/gutenberg/pull/300

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,13 @@ as of 2.0.0.
 
 ### Breaking
 
-- Remove any optimization logic + use of S3 cache (#300)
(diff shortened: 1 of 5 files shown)


rgaudin (approved):
LGTM

PR #326, opened 2025-10-16
Restore support for ZIM without full-text index + fix recent regressions

benoit74:
When moving from zimwriterfs to zimscraperlib, we inadvertently dropped the support for ZIMs without full-text index.

This PR restore this functionality + adds a warning when attempting to create a ZIM with multiple languages and a full-text index.

This PR also fixes "is_front" parameters to ensure that only books cover page is added to the ZIM to ensure we get only one entry in title search index per book. Title is also fixed to be "human-readable".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ as of 2.0.0.
 - Split build logic in Dockerfile to separate dependencies layer from scraper code layer (#302)
 - Prepare structure for TranslateWiki (#312)
 - Remove « .html » extension (#166)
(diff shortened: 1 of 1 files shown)


rgaudin:
This was done on purpose as the rule then was to never create a zim without FT index.

It's only recently that limitations of the search engine lead to a change in that policy. Btw is this documented?

benoit74:
> It's only recently that limitations of the search engine lead to a change in that policy. Btw is this documented?

In gutenberg there is a clear warning now ; in terms of policy, I only realized it "now", and I doubt it is clearly documented anywhere

PR #377, opened 2025-12-09
Filter authors to only include those with books

### LINKED ISSUE #359: Authors without books should not make it to the ZIM
Command used:

```
gutenberg2zim --lang ko
```

We end up with the `Anonymous` and `Various` authors automatically pushed to the ZIM while these authors are "fake generic ones" which should make it to the ZIM only if they have corresponding books.

Image

VikramAditya33:
- Add _get_authors_with_books() helper function in export.py
- Use filtered authors in generate_json_files() and generate_noscript_pages()
- Fixes #359 : Authors without books should not make it to the ZIM


benoit74 (approved):
LGTM

PR #378, opened 2025-12-12
Fix Kiwix-JS Restricted Mode: Remove meta refresh, use simple link in…

### LINKED ISSUE #364: No-JS fallback UI is not loading into Kiwix-JS restricted mode
See https://github.com/kiwix/kiwix-js/issues/1403

VikramAditya33:
Following the recommendation from @Jaifroid, we implemented Option 1: removed the meta refresh tag entirely and replaced it with a simple link in the `<noscript>` block. This allows Kiwix-JS to properly intercept the link click using its existing `parseAnchorsJQuery()` mechanism.
Fixes #364

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -50,7 +50,7 @@
       </style>
       <div class="message">
         <h1>JavaScript is Disabled</h1>
-        <p>This application requires JavaScript to function. Please click the link below to access the No-JavaScript version:</p>
+        <p>The usual ZIM interface isn't available right now because JavaScript isn't running in your ZIM reader. You can still view the content by using the No-JavaScript version below:</p>
         <a href="./noscript/books.html">Go to No-JavaScript Version</a>
       </div>
     </noscript>


ui/index.html
```diff
@@ -51,8 +50,7 @@
       </style>
       <div class="message">
         <h1>JavaScript is Disabled</h1>
-        <p>This application requires JavaScript to function. You are being redirected to the No-JavaScript version.</p>
-        <p>If you are not redirected automatically, please click the link below:</p>
+        <p>This application requires JavaScript to function. Please click the link below to access the No-JavaScript version:</p>
```
benoit74: I realize this wording could be confusing to users.

What about "The usual ZIM interface isn’t available right now because JavaScript isn’t running in your ZIM reader. You can still view the content by using the No-JavaScript version below:"

benoit74 (changes requested):
One small remark on wording to enhance clarity to any user, including non-tech-savvy. Feel free to make it even better if you have something to suggest

PR #396, opened 2026-02-15
Ensure reasonable access to books when JavaScript is disabled

### LINKED ISSUE #345: Design and provide a reasonable access to books when Javascript is disabled
With the move to Vue.JS, scraper will make JS an even more mandatory requirement.

We need to implement a fallback UI for situations where JS is disabled. UI does not need to be fancy at all (we accept we need Javascript for that) but should provide a reasonable listing of all books in the ZIM, and direct access to their various formats.

It is important title search and full text search also continue to work even with Javascript disabled.

rawadhossain:
### Summary
Project already has `noscript` pages for books, authors, and navigation, but some entry points still redirected to Vue routes, which do not work when JavaScript is disabled. So, some search results and navigation paths were unusable for no-JS users.

This PR ensures consistent access to books without JavaScript by redirecting search index entries and book information links to the corresponding `noscript` pages, making search, browsing, and reading flows usable without JavaScript.

### Changes
- Update `add_index_entry()` to redirect to noscript pages
- Remove unused `vue_route` parameter
- Add "View in full interface" link to noscript navigation
- Update book infobox link to point to noscript book pages instead of Vue routes

### Outcome
- Title and full-text search work without JavaScript
- Books, authors, and formats remain accessible via noscript pages
- Book information links while reading also work without JavaScript
- No-JS access is now consistent across search, browsing, and reading flows

### DIFF
--- a/scraper/src/gutenberg2zim/export.py
+++ b/scraper/src/gutenberg2zim/export.py
@@ -575,16 +575,20 @@ def _lcc_shelf_to_preview(shelf_code: str) -> LCCShelfPreview:
     )
 
 
-def add_index_entry(title: str, content: str, fname: str, vue_route: str) -> None:
-    """Add a custom item to the ZIM index with HTML redirect to Vue.js route.
+def add_index_entry(title: str, content: str, fname: str) -> None:
+    """Add a custom item to the ZIM index with HTML redirect to no-JS fallback page.
+
+    Search results (title search, full-text) must work without JavaScript. Index
+    entries redirect to noscript pages so both JS and no-JS users can access
+    content when clicking search results.
 
     Args:
         title: Title for the index entry
         content: Content/description for search indexing
-        fname: Filename for the index entry (e.g., "book_12345")
-        vue_route: Vue.js route path (e.g., "book/12345")
+        fname: Filename for the index entry (e.g., "book_12345") - must match
+            the noscript page name (e.g., noscript/book_12345.html)
     """
-    redirect_url = f"../index.html#/{vue_route}"
+    redirect_url = f"../noscript/{fname}.html"
     html_content = (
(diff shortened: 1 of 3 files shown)


scraper/src/gutenberg2zim/templates/book_infobox.html
```diff
@@ -1,6 +1,6 @@
 <div id="gutenberg-infobox">
     <a 
-        href="./index.html#/book/{{ book.book_id }}" 
+        href="./noscript/book_{{ book.book_id }}.html"
```
benoit74: This infobox is used by the JS interface and the noJS interface.

Either we find a elegant way to detect where we should like, or we keep the JS interface link which is main scenario we wanna cover.

scraper/src/gutenberg2zim/export.py
```diff
@@ -575,16 +575,20 @@ def _lcc_shelf_to_preview(shelf_code: str) -> LCCShelfPreview:
...
     Args:
         title: Title for the index entry
         content: Content/description for search indexing
-        fname: Filename for the index entry (e.g., "book_12345")
-        vue_route: Vue.js route path (e.g., "book/12345")
+        fname: Filename for the index entry (e.g., "book_12345") - must match
+            the noscript page name (e.g., noscript/book_12345.html)
     """
-    redirect_url = f"../index.html#/{vue_route}"
+    redirect_url = f"../noscript/{fname}.html"
```
benoit74: Changing this is going to break the JS experience, this is not what we want

rawadhossain:
I understand your concern. I think I misunderstood the issue, the approach should've been to preserve the default JS experience and only provide fallback when JS is unavailable. But I replaced JS behavior with no-JS behavior which breaks the current setup. Anyway since it's already merged, it was good learning for me.

PR #408, opened 2026-03-05
Fix: Misleading infobox assets logs

VikramAditya33:
The Infobox assets are exported first to fail fast if there's an issue, and the logs now accurately reflect the order of operations.
 >PR #405 and #404 had incorrect placement of this log


benoit74:
LoL, this is when you are forced to realized you may be a bit more tired than you wanna accept ^^

Never mind, I have to admit I did a force push of a completely outdated branch this morning on another repo 🙄

PR #414, opened 2026-03-07
Fail scraper when RDF download fails

### LINKED ISSUE #355: Fail the scraper run when RDF fails to download
Currently, when an RDF fails to download scraper just writes a warning and move on with next file.

In current situation, all RDF files are expected to be present, so I assume it would be safe to fail the scrape when an RDF file fails to download since it indicates a probable upstream issue.

At least [last full successful run](https://farm.openzim.org/pipeline/ee1f7b21-6a43-4004-9fde-63f979224433) of gutenberg_mul_all indicate no issue when downloading RDF files. Some books have been ignored because they are not available in any format, but this is expected, these are audio books.

Problem of ignoring RDF download errors is what happened in https://farm.openzim.org/pipeline/7023df29-a93b-4148-bc3b-d708ae6be2bd : mirror went offline for 16 minutes and the scraper continued. Should I have not been notified by mirror owner, I would not have cancelled the run and we would have had a ZIM with some books missing (but probably still consistent).

Scraper should hence immediately stop when an RDF file fails to download (after backoff, to ignore small intermittent networking issues).

VikramAditya33:
Currently, when an RDF file fails to download, the scraper logs a warning and continues processing other books. This can lead to incomplete ZIM files if a mirror goes offline temporarily during a scrape run.

For example, in [pipeline run](https://farm.openzim.org/pipeline/7023df29-a93b-4148-bc3b-d708ae6be2bd), the mirror went offline for 16 minutes and the scraper continued, which would have resulted in a ZIM with missing books.

### Solution:
Changed `download_and_parse_book_rdf()` to let exceptions propagate instead of catching them and returning `None`
Fixes #355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ as of 2.0.0.
 - Convert to WEBP and optimize WEBP, JPG and PNG images (both HTML resources + cover image) (#375)
 - Optimize Docker build topology to reuse layers (#420)
 - Optimize ePub again (#376)
+- Fail scraper when RDF download fails (#414)
 
 ### Changed
 
--- a/scraper/src/gutenberg2zim/rdf.py
+++ b/scraper/src/gutenberg2zim/rdf.py
@@ -199,29 +199,25 @@ def download_and_parse_book_rdf(book_id: int, mirror_url: str) -> Book | None:
         mirror_url: The mirror URL (e.g., "https://gutenberg.mirror.driftle.ss")
 
     Returns:
-        Book object if successful, None if book couldn't be downloaded or parsed
+        Book object if successful, None only for expected unusable books
+        (no license, no title, etc.)
+
+    Raises:
+        requests.RequestException: If RDF download fails (retries handled by caller)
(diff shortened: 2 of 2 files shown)


VikramAditya33:
@benoit74 I think there's a misunderstanding about where the retry logic is located. You're right that `requests.get()` itself doesn't retry, but i think that the retry logic exists at a higher level:

The `process_book_inner()` function in book_processor.py (lines 86-92) has a `@backoff.on_exception` decorator that already retries any `requests.exceptions.RequestException` (including from the RDF download), also uses exponential backoff (3s, 6s, 12s, 24s), and runs for up to 30 seconds (typically 4-5 retries), also Logs warnings on each retry via `backoff_request_error_hdlr`

Since `download_and_parse_book_rdf()` is called inside `process_book_inner()`, any network errors from the RDF download will trigger this retry mechanism. This fix removed the try-except blocks that were probably preventing exceptions from reaching this decorator.

The current implementation may already provide the retry behavior you're asking for. Could you please double check the `@backoff.on_exception` decorator in `book_processor.py` ?

PR #459, opened 2026-04-09
Add a function to remove MARC notations

### LINKED ISSUE #372: Strange character appears after colon in title
Every time there is a colon (":") in the title of a book, the next character will be $b
See for instance https://browse.library.kiwix.org/viewer#gutenberg_en_lcc-a_2025-12/Home
Image

VikramAditya33:
### Behavior 
The function removes MARC notation (:$a, :$b, etc.) and replaces it with ' : '

Fixes #372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ as of 2.0.0.
 
 ### Fixed
 
+- Add a function to remove MARC notations (#372)
 - Deduplicate cover images by reusing HTML version (#437)
 - Prevent crash when all books fail to download (#416)
 - Add Friulian (fur) language code mapping (#415)
--- a/scraper/src/gutenberg2zim/rdf.py
+++ b/scraper/src/gutenberg2zim/rdf.py
@@ -1,3 +1,4 @@
+import re
 from pathlib import Path
 
 import requests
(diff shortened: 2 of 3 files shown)


VikramAditya33:
@benoit74 I did check some most frequent notations or subfield on https://www.loc.gov/marc/ (particularly in bibliographic MARC format) and they were found to be 'abcdgtv'. I think we can narrow down to use this field only for regex but i feel like for generic MARC‑notation cleanup keeping 'a–z' is fine. What do you say?

benoit74 (commented):
I don't think this is correct, MARC fields are not always using the semi-colon. Can you help me understand why you stated this (beyond issue example of course)?

I also think we should probably clear all fields of the RDF file, not only the title (even if this is where we found this so far).

Finally we need at least few basic unit tests to validate the `clean_marc_notation` logic.

benoit74:
Not sure gutenberg is using this field, but you can for instance have a look at https://www.loc.gov/marc/bibliographic/concise/bd041.html

VikramAditya33:
Regarding the list `abcdgtv` there wasn't actually a strict official frequency table from where I pulled this, I just thought and saw that maybe only these ones might be mostly populated from what I figured out from typical MARC 21 bibliographic practice and examples.
Like : 
- `$a` is almost always present as the main element in many fields (author, title, subject, etc.).
- `$b`, `$c`, `$d` are common in name and title fields (e.g., numeration, titles/names, dates).
- `$g` often holds “miscellaneous information” (notes, statements of responsibility variants).
- `$t` marks a work title (e.g., series title or the title of a work within a field).
- `$v` is heavily used for form subdivisions in subject fields (e.g., `$650 $v Poetry`).

Though it was just an experimental thought of me 🤣 

Regarding your point related to always using colons i agree on removing all fields and also adding tests.

benoit74 (changes requested):
Please add a CHANGELOG entry and I think we are good to go

VikramAditya33:
Done

benoit74 (approved):

PR #453, opened 2026-03-31
Fix language codes not recognized correctly

Aunirbhan:
### Problem

`get_zim_language_metadata()` misses ISO 639-3 codes like `gla`, `nap`, and `oji` because it only looks up 2-letter keys in `ISO_MATRIX`. They resolve to `None`, producing an empty language list, and the ZIM crashes.

### Solution
- Track unresolved codes and mention them through a warning.
- Raises a clear error pointing to `--zim-languages` when nothing resolves.
- Added `ISO_MATRIX_REV` fallback so codes already known as values (like `gd→gla`, `oj→oji`) resolve correctly.
- @benoit74 I also added `nap` to `ZIM_LANGUAGES_MAP` since you mentioned it on the issue thread.

### Before / After

**Before** — `get_zim_language_metadata()` on upstream:
```python
get_zim_language_metadata(["gla"], books)  →  []  # silent, no warning
get_zim_language_metadata(["nap"], books)  →  []
get_zim_language_metadata(["oji"], books)  →  []
get_zim_language_metadata(["myn"], books)  →  []
```

**After**:
```python
get_zim_language_metadata(["gla"], books)  →  ["gla"]
get_zim_language_metadata(["nap"], books)  →  ["nap"]
get_zim_language_metadata(["oji"], books)  →  ["oji"]

get_zim_language_metadata(["myn"], books)  →  []
# WARNING: "Could not resolve ZIM language metadata for: myn"

get_zim_language_metadata(["myn", "nai"], books)  →  []
# WARNING: "Could not resolve ZIM language metadata for: myn, nai"
# ValueError: "Cannot resolve language metadata for: myn, nai.
#              Use --zim-languages to override."
```

Passes `hatch run test:run`

Closes #337

PR #470, opened 2026-04-23
Fix language codes not recognized correctly

### LINKED ISSUE #337: Some language are not recognized correctly?
Why https://farm.openzim.org/pipeline/8802dd7b-deb8-4254-92ab-25d290e3117c failed?

Looks like `gla` has not been recognized as a valid code.

Aunirbhan:
Addresses feedback from #453. Closes #337.

**Problem**
`get_zim_language_metadata()` misses ISO 639-3 codes like `gla`, `nap`, and `oji` because it only looks up 2-letter keys in ISO_MATRIX. They resolve to None, producing an empty language list, and the ZIM crashes.

**Solution**

- Added `ISO_MATRIX_REV` fallback so codes already known as values like `gd -> gla` and `oj -> oji` can resolve correctly.
- Added `nap` to `ZIM_LANGUAGES_MAP` as discussed in the issue thread.
- Used `resolve_language()` helper to simplify
- Made unresolved-language a single line by routing it through `resolve_language().`
- Restored the single-line `language_counts` dict comprehension
- Changed scraper behavior to fail immediately with a clear `ValueError` pointing to `--zim-languages` whenever a code cannot be resolved
- Added tests in `test_zim_language.py.`
- Added a `CHANGELOG` entry

Passes `hatch run test:run` with the added tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ as of 2.0.0.
 
 ### Fixed
 
+- Fix language codes not recognized correctly (#337)
 - Add a function to remove MARC notations (#372)
 - Deduplicate cover images by reusing HTML version (#437)
 - Prevent crash when all books fail to download (#416)
--- a/scraper/src/gutenberg2zim/iso639.py
+++ b/scraper/src/gutenberg2zim/iso639.py
@@ -240,5 +240,6 @@ ZIM_LANGUAGES_MAP: dict[str, list[str]] = {
     "myn": [],
     "nai": [],
     "nah": ["nhe"],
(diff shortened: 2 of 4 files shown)


benoit74:
@Aunirbhan why starting a new PR? This only makes review way more difficult to me because I have to re-understand what I've already reviewed + your new changes. 

Normal workflow is to open on PR per issue (or group of very related issues) and stack commits until we reach agreement on the solution. And only then consider if commit history make sense for the future or if we can stash them, reorder them, etc ...

Please move your new commits on top of existing commits in your `fix/337-language-resolution` branch and close this PR. I will review new changes in #453

PR #487, opened 2026-06-04
Replace manually-translated language names with CLDR data

siebrand:
Language names were previously maintained as translatable strings in each locale file, requiring translators at translatewiki.net to translate ~60 language names per UI locale — duplicating work already done in the Unicode CLDR project.

This PR replaces that approach with build-time extraction from CLDR data (cldr-localenames-full), supplemented by @wikimedia/language-data for autonyms and RTL detection.

Changes:

- Add a generation script that extracts translated language names from CLDR at build time
- Add language-names.ts utility for locale-aware language name lookup with graceful fallbacks
- Refactor format-utils.ts and useFormatters.ts to use CLDR data instead of i18n translator
- Derive supportedLanguages dynamically from available locale files
- Remove languageNames section from all locale files

This eliminates ~1000 lines of manually-maintained translations and removes the translation burden from translatewiki.net contributors, while providing better coverage (52 locales with CLDR vs. partial manual translations).


ui/scripts/generate-language-names.js
```diff
@@ -0,0 +1,65 @@
...
+import { fileURLToPath } from 'node:url'
+import { createRequire } from 'node:module'
+
+const __dirname = dirname(fileURLToPath(import.meta.url))
+const require = createRequire(import.meta.url)
+
+const cldrBase = dirname(require.resolve('cldr-localenames-full/package.json'))
+
+// Language codes used in Gutenberg books
+const LANG_CODES = [
```
benoit74, line 22: I know this was here before you came, but this list is wrong / outdated. I suspect this was done because usage was unclear ... I don't know but looking at the code this should be fine.

Correct list (extracted right now from PG):

```
const LANG_CODES = [ 
  'af', 'ale', 'ang', 'ar', 'arp', 'bg', 'bgs', 'bo', 'br', 'brx',
  'ca', 'ceb', 'cs', 'csb', 'cy', 'da', 'de', 'el', 'en', 'enm', 
  'eo', 'es', 'et', 'fa', 'fi', 'fr', 'fur', 'fy', 'ga', 'gl', 
  'gla', 'grc', 'hai', 'he', 'hu', 'ia', 'ilo', 'is', 'it', 'iu', 
  'ja', 'kha', 'kld', 'ko', 'la', 'lt', 'mi', 'myn', 'nah', 'nai', 
  'nap', 'nav', 'nl', 'no', 'oc', 'oji', 'pl', 'pt', 'rmq', 'ro', 
  'ru', 'sa', 'sco', 'sl', 'sr', 'sv', 'te', 'tl', 'yi', 'zh' 
]
```

ui/src/plugins/i18n.ts
```diff
@@ -12,72 +13,38 @@ export type Language = {
...
-  { code: 'sv', display: 'Svenska', rtl: false },
-  { code: 'te', display: 'తెలుగు', rtl: false },
-  { code: 'tl', display: 'Tagalog', rtl: false },
-  { code: 'zh', display: '中文', rtl: false },
-  { code: 'zh-hans', display: '简体中文', rtl: false },
-  { code: 'zh-hant', display: '繁體中文', rtl: false }
-]
+// Overrides for codes not in @wikimedia/language-data or where the autonym
+// differs significantly from what we want to display
+const AUTONYM_OVERRIDES: Record<string, string> = {
```
benoit74, line 18: Given new list of PG languages, I suspect we do not need `mul` and `oj` anymore?

ui/src/utils/language-names.ts
```diff
@@ -0,0 +1,78 @@
+import languageData from '@wikimedia/language-data'
+import cldrNames from '@/generated/language-names.json'
+
+// Codes not covered by either CLDR or @wikimedia/language-data
+const FALLBACK_NAMES: Record<string, string> = {
+  myn: 'Mayan',
+  nah: 'Nahuatl',
+  skr: 'Saraiki'
```
benoit74: Given new list, `skr` is not needed anymore

ui/src/utils/language-names.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Map internal codes to CLDR lookup keys
+const CODE_TO_CLDR: Record<string, string> = {
```
benoit74: Is this the mapping from TW code (used as-is internally) to CLDR lookup key? I'm not sure what "internal codes" means, could be PG code or TW code from my understanding.

benoit74 (changes requested):
Impressive work, thanks a lot. Now that you've shown how we should do this, it is clear that original idea was pretty crappy 🤣 Much appreciated.

I've spotted mistakes in the original list which could help simplify the codebase, probably worth to consider now that I found them.

Please also add an entry in CHANGELOG.md with link to this PR.

ui/scripts/generate-language-names.js
siebrand: Done.

ui/src/plugins/i18n.ts
siebrand: Correct. Will be in update.

ui/src/utils/language-names.ts
siebrand: Will be updated.

ui/src/utils/language-names.ts
siebrand: Will add comment.

siebrand:
Thanks for the review! [ea275f3](https://github.com/openzim/gutenberg/pull/487/commits/ea275f37dc6ca200ea7c3f93010f945d652c7189) and [be1d106](https://github.com/openzim/gutenberg/pull/487/commits/be1d10681fd1fac8d737ea6ab9920b083a20576b) hopefully address the comments on the changes.

siebrand:
I think I addressed all issues you pointed out. I may have already fixed one test. I don't really know your test framework, so maybe you can help out there a bit?

siebrand:
I think all comments were addressed and I fixed all CI/CD issues that came up. I've also update the commits to only have two.

benoit74 (approved):
LGTM, thank you

PR #488, opened 2026-06-04
docs: use correct translatewiki.net project name

siebrand:
The project is called translatewiki.net, not TranslateWiki.


benoit74 (approved):
LGTM, thank you (and sorry for using wrong name)

PR #494, opened 2026-06-08
fix: add typography in theme.ts and refactor header and footer

VikramAditya33:
- Migrated existing typography to use from `theme.ts`
- Refactored header and footer
- Added 'no author in this letter' text


VikramAditya33:
@kelson42 This is developmental branch, I'll get a review from @benoit74 once I'm done with the changes required, both of us have agreed on this.

PR #515, opened 2026-07-17
Fix: Code quality, correctness, and performance improvements across scraper

VikramAditya33:
# Fix: Code quality, correctness, and performance improvements across scraper

### 1. Packaging and tooling fixes (`pyproject.toml`)

**Fixed console script entry point**
- Changed `gutenberg2zim = "gutenberg2zim:entrypoint.main"` to `gutenberg2zim = "gutenberg2zim.entrypoint:main"`
- The old form required `__init__.py` to expose `entrypoint`, which was fragile. The new form directly imports `main` from `entrypoint.py`, matching how `__main__.py` does it.

**Fixed project homepage URL**
- Changed `https://github.com/openzim/kolibri` to `https://github.com/openzim/gutenberg`

### 2. Download engine fixes (`download.py`)

**Fixed HTTP response leaks**
- Replaced bare `requests.get()` calls with `with requests.get(...)` context managers in `resource_exists()`, `download_book()`, and `download_book_cover()`.
- **Outcome:** Connections are properly closed, preventing file descriptor exhaustion under concurrent processing.

**Fixed ZIP validation**
- When a downloaded ZIP returns HTTP 200 but `handle_zipped_html()` fails to extract anything (corrupt or rejected archive), the code now continues to the next preferred type instead of breaking out of the loop.
- **Outcome:** Books with corrupt ZIPs now try fallback HTML formats instead of being silently added with no content.

**Fixed quadratic byte concatenation**
- Replaced `content_bytes += chunk` (O(n²)) with `io.BytesIO()` (O(n)) for streaming downloads.
- **Outcome:** Significantly faster and more memory-efficient downloads for large files (PDFs, HTML ZIPs).

### 3. CLI fixes (`entrypoint.py`)

**Fixed default UI dist path**
- Corrected parent traversal from 5 levels to 4 levels to reach the repo root where `ui/dist` lives.
- **Outcome:** Default `--ui-dist` works correctly without explicit override.

**Fixed invalid LCC shelf handling**
- Added validation so that `--lcc-shelves=INVALID` fails fast with a clear error instead of silently falling back to "all shelves".

**Fixed `--zim-file` path handling**
- Preserved the full path from `--zim-file` instead of stripping the directory. Absolute paths are used as-is; relative paths with directories are preserved; bare filenames are placed in `--output`.
- **Outcome:** `--zim-file /tmp/library.zim` now writes to `/tmp/library.zim` instead of `./output/library.zim`.

### 4. Export and ZIM assembly fixes (`export.py`, `zim.py`)

**Fixed XSS vulnerability in generated HTML**
- Enabled Jinja2 autoescaping (`select_autoescape`) for HTML/XML templates.
- Added `html.escape()` to `add_index_entry()` for title, content, and redirect URL.

**Fixed exception swallowing in export**
- Removed `try/except` blocks that swallowed exceptions in `export_book()` for companion HTML and image processing. Added `raise` to the format handling exception handler.

**Fixed EPUB mimetype ordering and compression (two places)**
- In `optimize_epub_bytes()` and `zip_epub()`: the `mimetype` entry is now written first and uncompressed (`ZIP_STORED`), per the EPUB specification.
- **Outcome:** EPUBs are now standards-compliant and work with strict readers (Apple Books, Adobe Digital Editions).

**Fixed author-book index performance**
- Built `author_id -> books` mapping once in `generate_json_files()` instead of O(authors × books) repeated scans.
- Extracted the logic into `_build_author_books_map()` helper and reused it in `generate_noscript_pages()`.
- **Outcome:** Faster JSON and No-JS generation for large catalogs.

**Fixed regex replacement in index.html title**
- Changed `re.sub()` replacement string to a callback function with `html.escape()`.

### 5. Data model fix (`models.py`)

**Fixed empty selection vs no filter**
- Changed `if only_books:` to `if only_books is not None:` and used `set(only_books)` for O(1) lookup.

### 6. URL resolution fix (`pg_archive_urls.py`)

**Fixed template placeholder mismatch**
- Changed `.format(id=...)` to `.format(book_id=...)` to match the `{book_id}` placeholder in `FILENAMES`.
- **Outcome:** `archive_url()` no longer raises `KeyError: 'book_id'`.

### 7. Metadata parsing fix (`rdf.py`)

**Fixed missing title filtering**
- Changed default missing title from `"- No Title -"` to `""`.
- The downstream check `not parser.title` now correctly catches books without real titles.

### 8. Utility fixes (`utils.py`)

**Fixed MD5 hashing of binary images**
- Changed `read_file(fpath)[0].encode("utf-8")` to `fpath.read_bytes()`.

**Fixed encoding detection order**
- Moved `chardet` before the `iso-8859-1` fallback. Latin-1 was previously tried first and always "succeeded" (never raises `UnicodeDecodeError`), making `chardet` unreachable.

**Removed undeclared `six` dependency**
- Replaced `six.text_type(v)` with `str(v)`.

### DIFF
--- a/scraper/pyproject.toml
+++ b/scraper/pyproject.toml
@@ -52,11 +52,11 @@ dev = [
 ]
 
 [project.urls]
-Homepage = "https://github.com/openzim/kolibri"
+Homepage = "https://github.com/openzim/gutenberg"
 Donate = "https://www.kiwix.org/en/support-us/"
 
 [project.scripts]
-gutenberg2zim = "gutenberg2zim:entrypoint.main"
+gutenberg2zim = "gutenberg2zim.entrypoint:main"
 gutenberg2zim-validate-i18n = "gutenberg2zim.scripts.validate_i18n:main"
 
 [tool.hatch.version]
--- a/scraper/src/gutenberg2zim/download.py
+++ b/scraper/src/gutenberg2zim/download.py
@@ -39,8 +39,8 @@ class BookContent:
 
 def resource_exists(url):
     try:
-        r = requests.get(url, stream=True, timeout=DEFAULT_HTTP_TIMEOUT)  # in seconds
-        return r.status_code == requests.codes.ok
+        with requests.get(url, stream=True, timeout=DEFAULT_HTTP_TIMEOUT) as r:
+            return r.status_code == requests.codes.ok
     except Exception as exc:
         logger.error(f"Exception occurred while testing {url}\n {exc}")
         return False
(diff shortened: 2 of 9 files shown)


VikramAditya33:
Replacing `content_bytes += chunk` with `io.BytesIO()` for downloads does have a slight improvement in speed

VikramAditya33:
Everything works as it was, this is no breaking change

benoit74 (approved):
Well done, quite a significant load of small issues 🤮 

I let you merge at your convenience (let's do this each time, so that you handle order of PR merging as you see fit to avoid conflicts).

Issue #422, opened 2026-03-11
Review current 4.x design and identify final UI/UX enhancements


benoit74:
We are now getting close to releasing 4.x Gutenberg version. 

@kelson42 and @Popolechien have repeatedly said that they would like to see some UI enhancements.

I feel like involving @siemsie on the final review would be wise.

This is meant to be a "parent" issue to discuss stuff, with sub-issues for any topic which needs to be improved before released.

We have few issues left to implement, but these are only long awaiting bug fixes, no UI change expected.

We can hence base this issue discussion on last ZIM built today: https://dev.library.kiwix.org/viewer#gutenberg_de_all_2026-03/

@siemsie @kelson42 @Popolechien can you please review the ZIM mentioned above and report what we need to do, and how (probably worth to open one issue per topic not already identified)?

I will also link sub-issues which are already filled and UX-oriented and for which I have no clue how we want to solve them (I've already ruled some of some out of 4.0 release, but we can always discuss that, especially if we have a very clear idea of what we want to implement, not only a description of the issue we wanna solve)

VikramAditya33:
<img width="680" height="863" alt="Image" src="https://github.com/user-attachments/assets/2663751b-dd6a-4432-9235-a244c47de968" />

benoit74:
I've discussed with @siemsie live and we agreed she will make a design proposal to fix topics raised here and other concerns we have. Please allow us few days to settle on something. @VikramAditya33 would you like to participate to the discussions we will have with @siemsie? We will probably do this live to have more lively interactions and feedbacks.

siemsie:
Hi everyone. So the last couple of weeks have enabled me to gather my thoughts a little to put down some questions with regards to this project. Benoit and I have already discussed the below and require further feedback from @kelson42 and @Popolechien. @rgaudin's perspective is also welcome.

**1. Is Kiwix purely a shell for offline content or does it add its own brand ?**
My understanding has always been from previous ZIM designs is that Kiwix applies an aesthetic to the offline content that users' recognise from the original source/websites, hence our choice to do the same for this project. 
After Vikram's first pass at it (ii), we presented a new version (iii) that visually resembles Project Gutenberg (i), I was not in the meeting, but the feedback it seems what to break away from the source design and design our own "library".

Which leads me to ask: 

**1.1 Can we/_should we_ rebrand another's website or respect their aesthetics?** 
[...]

PR #516, opened 2026-07-22
UI refinements

VikramAditya33:
### What's new

* Replaced the old 5-star rating/popularity indicator with a new 3-fire popularity indicator.
* Removed pagination entirely. Books now use infinite scrolling throughout.
* Added a grey background for the focused author.
* Authors are now sorted alphabetically by their last name.
* Removed the format display from the book grid, and book titles can now span up to 3 lines.
* Fixed the author carousel navigation. Previously, it wasn't possible to navigate left from the currently focused author.
* Reduced the border thickness of book cells from **2px** to **1.5px**, and introduced a shared constant for it. The same constant is now also used by the LCC sidebar.
* Updated the footer text.
* Reduced the thickness of the breadcrumb component and fixed it so its thickness remains consistent.
* Improved the alphabet filter: **0–9** now also includes titles/authors starting with non-`A–Z` characters, and accented characters (e.g. **É → E**) are normalized to their base character.

This resolves almost all of the remarks from #422. I hope @siemsie and @benoit74 are happy with these changes now.

**Video showcase:**

https://github.com/user-attachments/assets/3e29aef2-6ee6-402e-b78d-0b67e517a812


VikramAditya33:
Also, after this, I hope I finally get to work on my favorite part of Project Gutenberg. 😄

benoit74 (approved):
I feel like we are definitely moving in the right direction and most issues are indeed solved.

I realize having all issues solved at once makes it very hard for me to review these changes, and even hard to express when I doubt we've fixed the thing exactly as intended, since I'm not even sure what we've decided we wanna do.

@siemsie do you see something which must be fixed or should we consider these are correct fixes and we will fix anything left later, in dedicated issues? I feel like we've reached something working pretty well and I would not be ashamed to start creating ZIMs with this.

siemsie:
I'm happy to move ahead so as not to slow the project down.

VikramAditya33:
Sounds fair. I'm merging this one.

If you ever feel like you need my attention on anything, don't hesitate to reach out, either here or on Slack.

VikramAditya33:
@benoit74 I agree with your point about refactoring everything at once. That said, I feel like this will hopefully be the last time I have to do it. From here on, I hope we can move forward incrementally instead.

PR #521, opened 2026-07-31
Ref: Pipeline refactor & export split

VikramAditya33:
#### Created `core/download_engine.py`

```python
class DownloadEngine:
    def __init__(self, session: requests.Session, cache_dir: Path, ...):
        ...

    def download(self, request: DownloadRequest, dest: Path) -> DownloadResult:
        ...
```

Features:
- Persistent session
- Retry with backoff
- Disk cache by URL hash
- Streaming to file

#### Split `export.py`

Moved generic pieces:

- `core/rewriters/link_rewriter.py`: `replacablement_link`, internal link rewriting
- `core/rewriters/image_rewriter.py`: image src rewriting, WebP conversion
- `core/exporters/json_exporter.py`: `generate_json_files`
- `core/exporters/nojs_exporter.py`: `generate_noscript_pages`
- `core/exporters/ui_dist_exporter.py`: `export_ui_dist`

Kept Gutenberg-specific pieces in `sources/gutenberg/plugins.py`.

#### Created `core/pipeline.py`

```python
class Pipeline:
    def __init__(
        self,
        catalog: CatalogPort,
        metadata: MetadataPort,
        resolver: FormatResolverPort,
        rewriter: RewriterPort,
        mapper: CollectionMapperPort,
        store: WorkStore,
        zim: ZimAssembler,
        engine: DownloadEngine,
        config: ScrapeConfig,
    ):
        ...

    def run(self):
        refs = self.catalog.discover(...)
        works = self.metadata.fetch(refs)
        ...
```

#### Created `core/concurrency.py`

```python
def parallel_map(func, items, concurrency):
    with Pool(concurrency) as pool:
        return pool.imap_unordered(func, items, chunksize=1)
```

#### Split `entrypoint.py`

- `cli.py`: only `parse_args()` and `main()`
- `config.py`: `build_scrape_config()`
- `orchestrator.py`: `run_scrape(config)`

#### Removed deprecated modules

- Deleted old `models.py`, `shared.py`, `csv_catalog.py`, `rdf.py`, `pg_archive_urls.py`, `download.py`, `book_processor.py`, `export.py`, `zim.py`.

### You're required to reinstall with `pip install -e scraper --no-deps` to regenerate it to point at `gutenberg2zim.cli:main` because your `.venv` might have the stale console script from the previous install.


benoit74:
@VikramAditya33 if the problem you face is the classical back vs ruff competition issue, then you might be interested to know that we now recommend to drop black and use only ruff. You can see this in action in the bootstrap and apply same changes than https://github.com/openzim/_python-bootstrap/pull/56

VikramAditya33:
@benoit74 It's ready for a review now, I know it's a big one but it was supposed to be like this otherwise I couldn't really split this PR into two. 1-2 PRs are remaining now before we completely refactor gutenberg and can add new sources with ease.

VikramAditya33:
Re black vs ruff, I'll open a dedicated PR once we complete this refactor we're only 1-2 PRs away from that

VikramAditya33:
The next one I'll open will most probably remove all the legacy functions that are still intact

scraper/src/gutenberg2zim/core/exporters/json_exporter.py
```diff
@@ -0,0 +1,463 @@
+"""JSON files exporter for the frontend UI (moved from `gutenberg2zim.export`).
+
+Generates books.json, authors.json, LCC shelves JSON, config.json, the
+per-item detail JSON files, and the ZIM index entries pointing at the
```
benoit74, line 4: The index entries are not exactly JSON files, it feels a bit weird to have them in `json_exporter.py`. At the same time, I agree it is very convenient from a code perspective. 

Maybe we should simply find a better name, something more "large" like `companion_data_exporter.py` (which is pretty ugly tbh).

Or we should split these to a `search_items_exporter.py` (which is a better name than "index" which is a bit vague IMHO).

I feel like the split would be a cleaner approach, but I'm did not tried to implement it ^^

benoit74 (commented):
I really like the shape all this is taking. Well done.

I have a small concern, which is more a discussion than a blocker from my perspective

benoit74:
> Re black vs ruff, I'll open a dedicated PR once we complete this refactor we're only 1-2 PRs away from that

Perfect, I was just concerned the competition between black and ruff was making CI fail. No hurry if you are not trapped in this dead end then.

scraper/src/gutenberg2zim/core/exporters/json_exporter.py
VikramAditya33: Fixed this concern in #522 as this was on point and that now `IndexBuilder` owns the derivation

benoit74: LGTM

benoit74 (approved):

PR #522, opened 2026-08-10
Pure Source-agnostic layer and core cleanup

VikramAditya33:
#### Streaming downloads

In `core/download_engine.py`, streamed to disk:

```python
with requests.get(url, stream=True, timeout=timeout) as r:
    r.raise_for_status()
    with open(dest, "wb") as f:
        for chunk in r.iter_content(chunk_size=chunk_size):
            f.write(chunk)
```

#### Disk cache

Cached RDFs and downloaded formats by URL hash:

```python
cache_key = hashlib.sha256(url.encode()).hexdigest()
cache_path = cache_dir / cache_key
```

#### Persistent sessions

Created one `requests.Session` per worker:

```python
session = requests.Session()
session.mount("https://", HTTPAdapter(max_retries=3))
```

#### `IndexBuilder`

Created `core/index_builder.py`:

```python
class IndexBuilder:
    def __init__(self, store: WorkStore):
        ...

    def build(self) -> Indexes:
        by_author = defaultdict(list)
        by_collection = defaultdict(list)
        search_entries = []
        ...
```

#### Granular retry

Moved retry from `book_processor.py` to individual downloads:

```python
@backoff.on_exception(backoff.expo, requests.RequestException, max_time=30)
def download_with_retry(...):
    ...
```

#### Legacy cleanup: split `plugins.py`, removed legacy `Book`/`Author` and `adapters.py`

**This section is a hard gate: (adding any new source) must not start until it is complete.** No new source may inherit, import, or route around Gutenberg's legacy dataclasses. The core must be genuinely source-agnostic before a second source proves it.

Two pieces of work:

**A. Split `sources/gutenberg/plugins.py`.** All remaining `export.py` functions were relocated into `sources/gutenberg/plugins.py`, making it a god file (~650 lines) mixing three concerns. Split it alongside the `downloader.py` → `DownloadEngine` migration, since that migration reshapes the export flow:

- `sources/gutenberg/rewriter.py`: `GutenbergHtmlRewriter`, `update_html_for_static`, `transform_image_path`, `export_infobox_assets` (the actual `RewriterPort` implementation)
- `sources/gutenberg/exporter.py`: `export_book`, `handle_book_files` (writing epub/pdf/html into the ZIM)
- `sources/gutenberg/epub_optimize.py`: `optimize_content`, `optimize_epub_bytes`, `_optimize_epub_jpeg/_png`, `_process_epub_html/_ncx`
- `plugins.py` is then removed

**B. Removed the legacy model layer entirely.** `core/exporters/json_exporter.py` and `nojs_exporter.py` currently imported `work_to_book` from `sources/gutenberg/adapters.py` and `Author`/`Book` from `sources/gutenberg/models.py`. A `core/` → `sources/` dependency that breaks the layering. Migrated both exporters and the Jinja templates to consume `core.models.Work`/`Creator`/`CollectionRef` directly, then deleted:

- `sources/gutenberg/models.py` (legacy `Book`/`Author` dataclasses)
- `sources/gutenberg/adapters.py` (`book_to_work` / `work_to_book`)
- `_all_books()` and the duplicate full-catalog `work_to_book` conversions (one per `_get_authors_with_books` call in both exporters.

Besides this some other scraper issues were fixed and `registry.py` has been introduced too


VikramAditya33:
A cache directory is now created in your output directory with contains cached RDFs when you run scraper. The scraper is quite faster now with the new download engine and this cache. I ran the scraper for 50 books and the cache directory created is of ~874KB which is quite small but useful in re-scraping the books.

VikramAditya33:
UI still hardcodes PG at many places so the next step will be that I'll create a UI config for PG and OTL and replace hardcoded strings and add OTL as a source along with separating the locales for both PG and OTL as locales can't be shared between sources too

VikramAditya33:
@benoit74 I made a small mistake here that #521 was base and this PR and #525 were stacked on top of it. All these tests were passing when these were stacked but I just merged #521 and these tests are failing as the scope of passing tests were on top of #521. I should've waited for all three to be approved but I hurried a little here. I apologize for that but you can approve these without hesitation as everything here is correct and no tests were failing. I misunderstood the new pull request stack feature a bit.

scraper/src/gutenberg2zim/core/concurrency.py
```diff
@@ -1,14 +1,21 @@
...
 from multiprocessing.dummy import Pool
 
 
 def parallel_map[T](func: Callable[[T], None], items: Iterable[T], concurrency: int):
-    """Map func over items using a thread pool with `concurrency` workers"""
+    """Map func over items using a thread pool with `concurrency` workers.
+
+    Uses `imap_unordered` with chunksize 1: work is dispatched lazily (the
+    whole item list is not chunked up front), workers never wait on
+    ordering, and worker exceptions still propagate to the caller when the
```
benoit74: Nice, worker exceptions propagation is a must ; are they immediately propagated, like they (mostly) immediately interrupt the loop?

scraper/src/gutenberg2zim/core/pipeline.py
```diff
@@ -108,6 +108,10 @@ def __init__(
         add_lcc_shelves: bool,
         primary_color: str | None = None,
         secondary_color: str | None = None,
+        # human-readable source name, used in ZIM titles, search entries
+        # and No-JS page titles; the orchestrator passes
+        # `SourceProfile.display_name`
+        display_name: str,
```
benoit74, line 114: `display_name` is quite obvious here, but then it is quite ambiguous in other places ; we don't really know what this is a display name for. Can you please find something more explicit?

benoit74 (changes requested):
@VikramAditya33, you did stacking right. Problem is `ValueError: Readme path must be within the project directory: ../README.md` ; something has changed and now refuse README from parent folder ; I saw that in other scrapers as well, I suspect this is a recent change in hatchling or something like that

Please investigate what proper solution should be (maybe only a matter of switching a flag saying we allow to use parent folder README).

Regarding the cache directory, I have some remarks. I get this is convenient for testing, but this will not be used on Zimfarm and most Docker image user who start from a fresh folder everytime. When doing a full ZIM, this will grow quite big without much added value. I feel like we should rather have a new CLI setting like `--rdf-cache` or `--cache-dir` or `--cache-folder` (or something like that) which allows to specify which folder to use (users might be interesting in storing that in another place than output depending on their disk layout). And by default, when CLI setting is not passed we do not cache RDF files. WDYT?

scraper/src/gutenberg2zim/core/concurrency.py
VikramAditya33: A failed book does not stop the full scrape: after retries, we log it and continue with the remaining books. This is intentional, since external links can be broken.  

If an error is not handled by the caller, `parallel_map` does stop and propagate it. Some already-running workers may finish first.

I’ll clarify the `parallel_map` docstring so it does not imply that the pipeline itself fails fast on individual-book errors.

scraper/src/gutenberg2zim/core/pipeline.py
VikramAditya33: Good point I’ll rename it to source_display_name throughout so it’s clear this is the human readable name of the source, such as Project Gutenberg or Open Textbook Library.

VikramAditya33: will fix this in #525

VikramAditya33:
@benoit74 The CI is passing now it was a Readme.md error. Besides this yes, I had the same concern while implementing cache directory especially for Zimfarm and full builds, where an RDF cache could grow large.
I agree that caching should be opt-in via a CLI option such as --cache-dir, allowing users to choose a suitable location for their disk layout. When it is not provided, we should avoid persisting RDF files. Do you want me to fix this here or on #525 ? It would be easier for me to work on #525 rather than here as I would have no trouble in rebasing and fixing merge conflicts.

scraper/src/gutenberg2zim/core/pipeline.py
VikramAditya33: refactored in #525 Implemented opt-in caching via: `--cache-dir=/path/to/cache`

VikramAditya33:
@benoit74 The CI happy Readme remains here only but I've fixed this behavior in #525

scraper/src/gutenberg2zim/core/concurrency.py
benoit74: This is indeed exactly the intended behavior

benoit74 (approved):
LGTM ; please merge and fix remaining issues in a later PR (no need to do that immediately in 525, could be a next one to keep topics separated).

Issue #95, opened 2019-11-12
Consider a better support of ZIM files without HTML


kelson42:
I think we should maybe consider a better support of ZIM files without HTML. The reasons are:
* This would save disk usage (I estimate a reduction of 30% of the whole size)
* Make easier the integration of third party ebook sources which don't (easily) provide HTML version of their content

Currently I see to big reasons to keep the HTML versions:
1 - Full text engine applying to HTML only
2 - Ability to directly see the content

These two things might be fixed with:
1 - Support ability to fulltext index EPUBs (relatively easy)
2 - Providing readers for multiple platforms within the ZIM... even maybe a pure Web Epub reader?

kelson42:
@gitenjar This ticket might be of interest for you

Popolechien:
Sure, sounds good but what would the final output look like compared to what we have now?

kelson42:
> Sure, sounds good but what would the final output look like compared to what we have now?

@Popolechien Same without the book in HTML directly usable from the browser, in place we would have an info page explaining how to read the EPUB file from the Browsers, mobile, computer, etc...

Popolechien:
@kelson42  Well, if the idea is to save some space, how about offering either Gutenberg epubs _or_ Gutenberg HTML and hope that people would know the difference? Much like we have Wikipedia with or without images, in a way.

kelson42:
> @kelson42 Well, if the idea is to save some space, how about offering either Gutenberg epubs _or_ Gutenberg HTML and hope that people would know the difference? Much like we have Wikipedia with or without images, in a way.

This might be done, can be already one, but this is not the point of the ticket which is about providing a better support without HTML.

Popolechien:
> we should provide one with HTML and one with EPUB and people can only have one or the either and live with that.

Yes.

kelson42:
@Popolechien To me this would be a fallback solution. But I believe we might be able to solve the problem properly.

We could be able to solve (2) in an even better manner by using a pure javascript EPUB reader (so for the end user) it would be a similar experience as having the HTML in the ZIM file. We could for example use https://github.com/futurepress/epub.js/

eshellman:
only tricky thing deploying epub.js is overcoming same-origin javascript issues, but you probably are experienced with that

kelson42:
[...]

PR #525, opened 2026-08-14
Add Open Textbook Library as a new source

### LINKED ISSUE #465: Current architecture planning and refactoring
The current architecture of Gutenberg2zim suffers from from tight coupling and poor testability as described by @benoit74 here: https://github.com/openzim/gutenberg/issues/422#issuecomment-4108783273.

From what I've realized that there is also a new architecture plan going on to introduce new ingestions/scrapers, example: https://github.com/openzim/zim-requests/issues/1258 but this is another discussion which is going on here: https://github.com/openzim/gutenberg/issues/94.

Out of these two what are we leaning towards mostly? From what I think is that Gutenberg2zim has grown organically with time and now it probably requires an architecture refactor which would probably enhance testability and also with loosely coupled architecture it would significantly enhance performance and maintainability too. 

Introducing new scrapers would require to extend this repo in which I probably think (Monorepo + Common Library) would be a good choice for this but again this requires a distinct discussion which is probably going on and also what I think is that building multi-source on top of the current architecture would compound existing problems.

We can keep this as a parent issue for other architecture related issues that will be opened in near future.

VikramAditya33:
### Link to the ZIM

Drive: <https://drive.google.com/drive/folders/1p7e8PRCF20bcRT2wpUsMga440_q80drf?usp=sharing>

This PR adds Open Textbook Library (OTL) as a second source alongside Project Gutenberg.

OTL discovery uses its CSV export:

```text
https://open.umn.edu/opentextbooks/download.csv
```

Full per-book metadata comes from the OTL JSON API:

```text
https://open.umn.edu/opentextbooks/textbooks/{id}.json
```

### Scraping OTL

```bash
gutenberg2zim \
  --source opentextbooks \
  --output ./output \
  --zim-file ./output/opentextbooks.zim
```

Select a positional range from the eligible CSV catalog:

```bash
--books 1-50
```

Select exact OTL record IDs:

```bash
--otl-ids 28,126,127
```

Filter by subject:

```bash
--subjects "Mathematics,Computer Science"
```

### Optional caching

Caching is opt-in and never stored in the output directory. Pass the same cache directory on later runs to reuse OTL catalog and metadata data, invalid URL records, and Gutenberg RDF metadata:

```bash
--cache-dir ./cache/otl
```

Refresh the OTL CSV catalog only when caching is enabled:

```bash
gutenberg2zim \
  --source opentextbooks \
  --cache-dir ./cache/otl \
  --refresh-catalog
```

Without `--cache-dir`, OTL fetches its CSV and metadata in memory for that run only.

OTL downloads validated PDF and EPUB editions when available, with complete offline HTML editions as a fallback. It includes covers, subjects, and review-based flame ranking.

### Readers

Added built-in readers to the UI:

- EPUB reader based on Foliate, with navigation and a clickable table of contents.
- PDF reader with standard document controls.
- Offline HTML editions retain navigation, book information, available-format downloads, and scroll-to-top controls.

### Locales

Locales now use one Translatewiki-compatible file per language:

```text
locales/en.json
locales/fr.json
...
```

Each file uses top-level namespaces:

```text
common
gutenberg
opentextbooks
```

The UI merges `common` with the active source namespace, keeping shared messages source-agnostic and source wording isolated.

### Source-agnostic architecture

Core models, pipelines, exporters, UI collection components, and locale loading are source-agnostic. Source-specific behavior is isolated behind source profiles and adapters, providing a reusable foundation for additional sources.


VikramAditya33:
```
.venv/bin/gutenberg2zim --output ./output --zim-file ./output/otl_1-50.zim --books 1-50 --overwrite --source opentextbooks
```
I use this command most commonly for me

VikramAditya33:
Another thing I need to confirm is that I didn't apply epub compression that is being used for PG on OTL because it removes PG boilerplate and edits its NCX/HTML. That would be wrong for publisher-provided OTL books.

For it to work on OTL, we need to build a source-neutral EPUB optimizer.

VikramAditya33:
@benoit74 this one is passing tests because now base is #522 but I think the tests will fail for this too if we merge #522 alone and leave this as it is :rofl: I really misunderstood merge stack.

scraper/pyproject.toml
```diff
@@ -8,7 +8,7 @@ authors = [{ name = "Kiwix", email = "dev@kiwix.org" }]
 keywords = ["kiwix", "zim", "offline", "gutenberg"]
 requires-python = ">=3.14,<3.15"
 description = "Make ZIM file from Gutenberg books"
-readme = "../README.md"
```
benoit74: This is what "saves" the CI to pass. But then we are not using the proper README anymore ...

benoit74 (changes requested):
Very good !

Unfortunately we cannot have multiple locales folders on Translatewiki, or at least it would be too much pain.

I propose we rather still have only one folder and one file per language, and solve the `common` vs `gutenberg` vs `otl` "simply" by moving these at "top level" keys in the JSON file.

benoit74:
Important to note that this partly solved #95 by providing the ePub reader. We still miss the capability to index ePubs, but this is for the scraperlib (https://github.com/openzim/python-scraperlib/issues/333)

scraper/pyproject.toml
VikramAditya33: Fixed it the way you intended, the package now publishes the actual top-level README.md. `scraper/hatch_build.py` supplies the root README contents as Hatchling’s dynamic package readme

VikramAditya33:
@benoit74 Oh my bad I didn't know about this limitation translatewiki had

VikramAditya33:
@benoit74 I've structured generic UI keys like : 

```
{
  "common": {
    "nav": { "collections": "Collections" },
    "book": { "collection": "Collection" }
  },
  "gutenberg": {
    "nav": { "collections": "Shelves" }
  },
  "opentextbooks": {
    "nav": { "collections": "Subjects" }
  }
}
```

They all live under one folder now

VikramAditya33:
@benoit74 Implemented opt-in caching via:
`--cache-dir=/path/to/cache`

VikramAditya33:
Do I merge this and #522 now @benoit74 ?

benoit74:
Once I approved you can merge ; unless I request a minor change, but if I don't say anything and just approve, we are fine and you can merge. So you can merge #522, and wait for my feedback ~~merge~~ before merging this one

VikramAditya33:
@benoit74 merged #522 as per your request. This one is ready for your review

VikramAditya33:
also edited the PR description according to the current behaviour

README.md
```diff
@@ -86,9 +86,17 @@ Note: You'll need to install system dependencies (zim-tools) separately. See [CO
...
 --mirror-url=<url>                   Custom Gutenberg mirror URL
 --output=<folder>                    Output folder (default: ./output)
+--cache-dir=<folder>                 Optional persistent metadata and catalog cache
 --debug                              Enable verbose output
 ```
 
+### Caching
+
+Caching is opt-in. Pass `--cache-dir` to persist source metadata and catalog
+data outside the ZIM output directory. To reuse cached data in a later run,
```
benoit74: The "outside the ZIM output directory" part is confusing. I can understand it as "you must not use same directory as ZIM output directory", which in fact is not a concern AFAIK. And I can also understand it as "if you do not pass this setting, the scraper will persist data in the ZIM output directory", which is not scraper behavior.

scraper/src/gutenberg2zim/sources/gutenberg/cli.py
```diff
@@ -0,0 +1,72 @@
...
+    "--lcc-shelves=<shelves>         Comma-separated LCC shelf codes to include "
+    "(e.g., P,PR,Q). Use 'all' for every shelf"
+)
+
+
+def parse_options(arguments: dict[str, Any]) -> dict[str, Any]:
+    for option in ("--subjects", "--otl-ids", "--list-subjects", "--refresh-catalog"):
+        if arguments.get(option):
+            critical_error(f"{option} belongs to --source opentextbooks")
+    value = arguments.get("--lcc-shelves")
```
benoit74: `value` is very unclear, `lcc_shelves_arg` or something like this would be way more explicit

scraper/src/gutenberg2zim/sources/gutenberg/cli.py
```diff
@@ -0,0 +1,72 @@
...
+}
+
+OPTIONS = (
+    "--lcc-shelves=<shelves>         Comma-separated LCC shelf codes to include "
+    "(e.g., P,PR,Q). Use 'all' for every shelf"
+)
+
+
+def parse_options(arguments: dict[str, Any]) -> dict[str, Any]:
+    for option in ("--subjects", "--otl-ids", "--list-subjects", "--refresh-catalog"):
```
benoit74, line 56: This is not sustainable in the long term. We cannot maintain in each source the list of other sources reserved arguments. No need to solve this now (it works fine and is totally OK for two sources) but please open an issue so that we rework this part at some point: each source can expose its custom arguments, and the core can handle raising critical errors when it sees an argument which does not belong to current source. Or maybe it could be even simpler to rework all this args parsing logic, does not look very modern tbh.

ui/scripts/generate-language-names.js
```diff
@@ -1,7 +1,7 @@
 /* global Buffer, console */
 /**
  * Extracts translated language names from CLDR data for use at runtime.
- * Only includes the language codes used in Gutenberg and only for UI locales
+ * Only includes catalog language codes used by the bundled sources and UI locales
```
benoit74: I'm a bit lost here. Isn't all this mostly valid on for Gutenberg which chose to use ISO639-3? And isn't the "CATALOG_LANGUAGE_CODES" only a list valid for Gutenberg? Or is it CLDR equivalent of files present in `locales`?

benoit74 (changes requested):
Impressive work, thank you very much. Some feedback and interrogations below, but we are close to merge IMHO.

README.md
VikramAditya33: I'll fix the wording

scraper/src/gutenberg2zim/sources/gutenberg/cli.py
VikramAditya33: lcc_shelves_arg seems the right choice

scraper/src/gutenberg2zim/sources/gutenberg/cli.py
VikramAditya33: oh yea I forgot to centralize this and just implemented a temporary workaround, indeed it is painful to manage. I've opened up the issue : #529

ui/scripts/generate-language-names.js
VikramAditya33: ahh how did I miss this. Thanks for pointing this out and yes indeed it is gutenberg specific only, OTL doesn't add to it. I made a mistake here

VikramAditya33:
@benoit74 Thanks for pointing these mistakes out. I've resolved all the issues that you mentioned and opened the new issue regarding cli args #529 it is.

ui/scripts/generate-language-names.js
benoit74: But isn't ~`generated/languages.json`~ `generated/language-names.json` used by all sources ATM? I'm still confused

VikramAditya33: No because OTL's CSV discovery currently assigns every entry `en`, and JSON metadata only normalizes it. Though this list was not made source agnostic as I guess would be better if a future source contributes other language codes, then the generator would explicitly combine per source list instead of this static list we have right now.

VikramAditya33: What do you say?

benoit74: OK, this is indeed OK for now, let's wait for next multi-lingual source to appear and we will adapt as needed.

benoit74 (approved):
LGTM, I let you merge

Nikerabbit:
The `@metadata` key in the translation files must be at the top level.

VikramAditya33:
@Nikerabbit thanks for letting this know, I'll take a look

VikramAditya33:
@Nikerabbit Fixed in #534

PR #532, opened 2026-08-28
Centralize source-specific CLI option validation

### LINKED ISSUE #528: Add short form name aliases for different sources
#527 says to add "Bibliothèque numérique romande" as a new source, but writing this whole name in CLI is a headache. 

Instead we can map sources to their short form, basically to the initials of their letters (in capital): 

Project Gutenberg = PG
Open Textbook Library = OTL
Bibliothèque numérique romande = BNR

etc... for future sources

### LINKED ISSUE #529: Centralize validation of source-specific CLI options
Source-specific CLI parsers currently need to recognize options belonging to other sources in order to reject them. This works with the current two sources, but does not scale as new sources are added.

Rework CLI argument handling so each source declares its custom options and the core validates whether supplied source-specific options belong to the active source. Consider modernizing the broader parsing flow while doing so.

Current workaround: `sources/gutenberg/cli.py` explicitly rejects OTL-specific options.

VikramAditya33:
## Improve source-specific CLI options

Source-specific CLI options are now declared by their owning source and validated centrally. This removes cross-source option checks from individual source parsers and makes additional sources easier to add.

The CLI now rejects options that belong to a different selected source, including explicitly empty values. Duplicate custom option declarations are rejected during source registration.

Short source aliases are also supported:

```bash
gutenberg2zim --source PG
gutenberg2zim --source OTL
```

Fixes #529 
Fixes #528

### DIFF
--- a/scraper/src/gutenberg2zim/cli.py
+++ b/scraper/src/gutenberg2zim/cli.py
@@ -12,7 +12,11 @@ from gutenberg2zim.constants import VERSION
 from gutenberg2zim.orchestrator import run_scrape
 from gutenberg2zim.sources.registry import SOURCES
 
-source_options_help = "\n".join(profile.cli_options for profile in SOURCES.values())
+source_options_help = "\n".join(
+    option_help
+    for profile in SOURCES.values()
+    for option_help in profile.cli_options.values()
+)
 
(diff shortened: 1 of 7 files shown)


benoit74 (approved):
LGTM

PR #537, opened 2026-08-31
Expose multi-source options in Zimfarm recipes

### LINKED ISSUE #536: Update offliner-definition.json
Recent changes in CLI arguments linked to OTL integration, e.g. the new `--source` parameter, have not been updated in https://github.com/openzim/gutenberg/blob/main/offliner-definition.json (which is the file used by Zimfarm to display recipe options).

ATM it is not possible to have something "dynamic" which show/hide some argument based on another argument value, so we will live without it (expose both Gutenberg specific and OTL specific arguments).

Note that it is possible (and recommended for `source`) to define enums: https://github.com/openzim/mwoffliner/blob/6f648ac9227625a7fec92d89d03ec665044b24f5/offliner-definition.json#L110-L122

@VikramAditya33 I've assigned the issue to you because you are in the best position to tackle it, but feel free to speak up if you don't have time for that

VikramAditya33:
Updates `offliner-definition.json` so Zimfarm can configure the supported book sources.

- Adds a `source` dropdown with Project Gutenberg and Open Textbook Library choices.
- Changes formats to a multi-select enum.
- Exposes OTL-specific `subjects` and `otl_ids` filters.
- Clearly labels Gutenberg-only LCC shelf filtering.
- Updates shared option descriptions for source-neutral behavior.

Zimfarm cannot conditionally show options based on the selected source, so both source-specific filters remain visible and explain their valid source.

Administrative CLI actions (`list_subjects`, `refresh_catalog`, and `cache_dir`) are intentionally excluded because they do not define a ZIM-building recipe.

Adds regression tests for the recipe definition.

Tests: `pytest scraper/tests -q`

Fixes #536

### DIFF
--- a/offliner-definition.json
+++ b/offliner-definition.json
@@ -3,6 +3,22 @@
   "stdOutput": true,
   "stdStats": true,
   "flags": {
+    "source": {
+      "type": "string-enum",
+      "required": false,
+      "title": "Source",
+      "description": "Book source to scrape. Source-specific options are shown below; do not combine options from different sources.",
+      "choices": [
+        {
+          "title": "Project Gutenberg",
+          "value": "gutenberg"
+        },
+        {
+          "title": "Open Textbook Library",
+          "value": "opentextbooks"
+        }
+      ]
+    },
     "languages": {
       "type": "string",
       "required": false,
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM

PR #585, opened 2026-09-21
fix: make home.mostDownloaded label specific to each source

### LINKED ISSUE #584: `home.mostDownloaded` label must be specific to each source
Currently, the `home.mostDownloaded` label is common to all sources.

This is wrong because on OTL we do not have downloads but reviews.

We should hence move this label to source-specific section of the locales files, and set proper value on OTL.

@VikramAditya33 does this makes sense to you? Would you like to work on this?

VikramAditya33:
fix #584


benoit74 (changes requested):
We should change `mostDownloaded` key to `mostPopular` for clarity IMHO

VikramAditya33:
Oh makes sense

benoit74 (approved):
LGTM

PR #588, opened 2026-09-21
CI: use zim-tools Docker image instead of downloading it

### LINKED ISSUE #586: CI should not download zim-tools but rather use Docker image
We should prefer to use Docker image at https://github.com/openzim/zim-tools/pkgs/container/zim-tools instead of downloading zim-tools from Kiwix server, this would give better speeds.

pawank925:
Fixes #586

## Changes

- Remove the "Install zimtools" step (`wget`/`tar` download) from `Tests.yaml` and `DailyTests.yaml`
- Run `zimcheck` through the `ghcr.io/openzim/zim-tools:3.8.0` container image, mounting the `output` directory the same way the other Docker steps in those workflows do


pawank925:
@VikramAditya33

benoit74 (approved):
Thank you very much @pawank925

PR #589, opened 2026-09-21
Gutenberg: deduplicate cover images bundled in HTML zip

### LINKED ISSUE #581: Gutenberg: deduplicate cover images
When using Gutenberg scraper, we end-up quite often with two times the same image in the ZIM file: once from the HTML format and once from the fact that we download the cover image anyway for UI display.

We might want to consider this is wasted space and try considering we could deduplicate these images, choosing the best one.

Since this deduplication is gonna be complex (it is often not the exact same image from binary perspective, not same size), we should first evaluate how much size it would save in the ZIM

pawank925:
closes #581

## Changes

 In `handle_book_files`, when no cover was detected through the HTML `<link rel="icon">` href, also detect the cover asset bundled by PG in the zip (`cover.jpg`/`cover.png`/`{id}-cover.png`), so `export_book` aliases `covers/{book_id}_cover_image.webp` to the already-stored HTML asset instead of downloading the mirror cover.
 The bundled cover is usually the full-size picture, so this also selects the best version, and the mirror download is skipped entirely.
Add unit tests covering bundling detection, aliasing, the mirror-download fallback when no cover is bundled, and the existing icon-link path (regression).

## Tests

`pytest tests/` passes (193 tests), `black`/`ruff` clean, `pyright` reports no errors on the touched modules.


pawank925:
@VikramAditya33

benoit74 (changes requested):
Thank you, I like it!

One correctness issue worth fixing before merge, plus a minor nit.

### Cover detection runs before optimization, so a failed optimization leaves a dangling cover reference

In `_add_asset_file`, `_detect_html_cover` (which sets `work.extra["html_cover_path"]`) runs *before* `optimize_content`, which can raise:

```python
def _add_asset_file(work: Work, filename: str, file_content: bytes, assembler):
    output_filename = ImageProcessor.get_output_filename(filename)
    _detect_html_cover(work, filename, output_filename)   # sets html_cover_path
    optimized_file_content = optimize_content(work, filename, file_content)  # can raise
    assembler.add_item_for(...)
```

**Failure scenario:** a bundled cover image that PIL can't decode/re-encode (corrupt bytes, unsupported CMYK JPEG, etc.) causes `optimize_content` to raise *after* `html_cover_path` has already been set. The exception is caught in `_add_associated_files`, logged, and the item is never added to the ZIM — but `export_book` still sees `html_cover_path` truthy, so it aliases to a ZIM entry that doesn't exist **and** skips the mirror-download fallback entirely. The book ends up with no working cover at all, which is worse than the pre-refactor behavior (a failed optimization simply left the cover undetected and the mirror fallback still ran).

**Suggested fix** — do the work that can fail first, and only mark the file as "the cover" once it's actually been optimized and added:

```python
def _add_asset_file(work: Work, filename: str, file_content: bytes, assembler):
    output_filename = ImageProcessor.get_output_filename(filename)
    optimized_file_content = optimize_content(work, filename, file_content)
    assembler.add_item_for(
        path=output_filename,
        content=optimized_file_content,
        is_front=False,
    )
    _detect_html_cover(work, filename, output_filename)
```

This way, if optimization (or `add_item_for`) fails, the exception propagates before `html_cover_path` is touched, and the mirror-download fallback engages exactly as it did pre-refactor. Would be good to add a regression test that exercises `optimize_content` raising for a bundled-cover file and asserts the mirror fallback still runs.

###  Nit: companion-HTML and asset-file errors now share one log message

`_add_associated_files` used to log a distinct `Exception while handling companion HTML: {e}` message; both branches now share the generic `Exception while handling file {filename}: {e}`. Pure log-message regression (no functional impact), but it makes it slightly harder to tell from logs alone whether an HTML-rewrite or an image/asset failure occurred. Worth a one-line fix if convenient, not a blocker.

benoit74:
@pawank925 thank you. Next time, please avoid force push and merging from main to feature branch during the review. We prefer to rebase and only do that once review is completed. This simplifies future maintenance with a simple commit tree and avoids noise during review. Exceptions are of course possible when you need changes which already arrived in `main` (but then prefer a rebase).

benoit74 (approved):
LGTM

PR #591, opened 2026-09-22
Separate raw popularity from flame ratings

### LINKED ISSUE #583: Work (book) sorting is wrong and there is a confusion between popularity and downloads
Currently, each work (book) has a `popularity` and a `primary_metric` value attached:
- the `primary-metric` is the number of downloads in PG, number of reviews on OTL
- the `popularity` is the number of flames (0 to 3, integer only), which the scraper computes by binning works based on their `primary_metric` 

Sorting works by popularity in the UI is based on `popularity`, which is not exceptional since it is a discrete value, so sorting is a bit random inside a given bin. This is especially bad when we want the "best work" (i.e most downloaded on PG) to display. The consequence is that as of today the "most downloaded" work shown on home page in PG ZIMs is often not the most downloaded at all, just the first work in bin "3".

We will hence rename `primary-metric` to `popularity` and use that to sort works in the UI ; we will also rename `popularity` to `flames` for clarity. Both needs to be in work preview (only `flames` are sent for now).

@VikramAditya33 does it makes sense to you as well?

VikramAditya33:
## Separate raw popularity from flame ratings

Works now retain their source-specific raw popularity score:

- Project Gutenberg: download count
- Open Textbook Library: aggregate review score, with review count as a small tie-breaker

Flames are derived from that same score and remain a discrete 0–3 visual indicator.

The UI now sorts and selects featured books using raw popularity, so the featured work is genuinely the highest-ranked one even when multiple works share the same flame count. Raw popularity and flames are included in work previews.

Raw popularity values are no longer shown in book details or no-JS pages, since they are useful for ordering but do not represent a meaningful offline metric.

Adds regression coverage for selecting the highest-ranked work within the same flame bucket.

Fix #583


benoit74 (approved):
One minor finding: formatMetric in ui/src/utils/format-utils.ts is now dead code — its only caller was BookDetailInfo.vue's primaryMetric display block, which this diff removed, but the function and its dedicated test in format-utils.spec.ts were left behind. Worth deleting both before merge.

@VikramAditya33 I let you clean that and merge on your own

VikramAditya33:
nice catch. I'll refactor and merge

PR #594, opened 2026-09-22
Fix underline bleeding into space before collection count

### LINKED ISSUE #592: Shelves: link should not contain the space
In the left part of the shelves display, 

Image

We have a very small glitch:

Image

The underline is covering the space where we would like to have no underline on the space:

Image

Clearly a very small visual glitch, but since we saw it, I fill the issue

luizestevestata-prog:
Fixes #592.

## Problem

On the collection sidebar, when a collection link is active it gets `text-decoration: underline`. The label text and the `(count)` span were separated in the template by whitespace, which Vue rendered as a real space character living directly inside the underlined `<span>`. Since that space wasn't part of the `.collection-sidebar__count` inline-block, the underline bled into the gap right before the count.

## Fix

Moved that space to be the leading character *inside* `.collection-sidebar__count` instead of a sibling text node. Since `.collection-sidebar__count` is `display: inline-block`, ancestor text-decoration doesn't propagate into it — so the underline now stops right after the label, matching the "fixed" screenshot in the issue. No visual spacing change, only the underline extent.

Applied the same fix to both the "All collections" row and the per-collection rows, since both share this markup pattern.

## Testing

Could not run the full dev server locally (needs a live backend for collection data), so I verified the underlying CSS/whitespace behavior with an isolated HTML reproduction of the same markup and styles, confirming the underline no longer extends past the label text once the space lives inside the inline-block count span.


benoit74:
This is not working, we have no more space: 
image

See https://github.com/openzim/gutenberg/blob/main/CONTRIBUTING.md#3-developing-the-vuejs-ui-with-real-data for solution to test with real data without needing to download the whole things every time

benoit74 (changes requested):
Please fix the bug

PR #595, opened 2026-09-22
feat: implement the reusable EPUB optimizer

### LINKED ISSUE #552: Introduce a source-agnostic EPUB and image optimization engine
Move reusable EPUB archive/image optimization into `core/` so both Gutenberg
and OTL can use it.

Scope:
- preserve EPUB ZIP invariants (`mimetype` first/uncompressed, member order,
  timestamps, attributes, duplicate members);
- optimize embedded JPEG/PNG only when smaller, without converting them to
  WebP;
- expose source-owned callbacks for document transforms.

Gutenberg must retain its own HTML/XHTML cleanup and NCX license removal.
OTL should use only the generic optimizer initially.

Do not add optimized-asset caching in this work.

VikramAditya33:
## Reuse EPUB optimization across sources

Moves EPUB archive and embedded-image optimization into a source-agnostic core module.

The core optimizer:

- preserves EPUB ZIP requirements: `mimetype` first and uncompressed, member order, timestamps, ZIP attributes, archive comments, and duplicate entries;
- optimizes embedded JPEG and PNG files only when the result is smaller;
- retains JPEG/PNG formats rather than converting them to WebP;
- safely retains the original image when optimization fails;
- accepts optional source-owned callbacks for HTML/XHTML/NCX document transforms;
- performs all work in memory, without optimized-asset caching.

Gutenberg now supplies its existing EPUB HTML/XHTML cleanup and NCX license-removal callback to the core optimizer. OTL now uses the generic optimizer without source-specific document transforms.

Adds coverage for ZIP invariants, duplicate members, metadata preservation, image-size decisions, failed image optimization, source document callbacks, and OTL EPUB integration.

Fix #552


benoit74 (commented):
Thank you very much, well done!

I failed to find anything worth to mention, but Claude review spotted two things I find worth to mention:

1. `core/epub_optimizer.py:96-98` — Unexpected `.gif/.webp` members inside an EPUB used to log `logger.warning(f"Unexpected {suffix} image in EPUB for book {work.id}: {name}")` (a signal of a corrupted/unusual source file). In _optimize_member, these suffixes now match neither `_IMAGE_OPTIMIZERS` nor `_DOCUMENT_SUFFIXES` and silently fall through to return data — no log at all, for both Gutenberg and OpenTextbooks.
2. `core/epub_optimizer.py:30-74` — The old code computed `optimized_size` vs `original_size` and logged a warning when optimization made the EPUB larger, or a debug line otherwise. That comparison is gone entirely in the new `optimize_epub_bytes` — no size-regression signal survives the reorg, and your own memory notes call this diagnostic out as useful for spotting real-world cases.

You did not mentioned reasons for dropping these. Given the relative fragility of this ePub optimization algorithm, I recommend that we keep these logs. WDYT?

VikramAditya33:
The warning part was gutenberg specific only though I will retain this this and the comparison one as well. Nice catch by claude

VikramAditya33:
Implemented both the findings PTAL

benoit74 (approved):
LGTM, thank you very much

Issue #593, opened 2026-09-22
Show number of author books in book details view


benoit74:
Add the number of books the author has in current archive, to avoid clicking on the name only to realize the author has only one book... kinda deceiving ^^

<img width="1265" height="590" alt="Image" src="https://github.com/user-attachments/assets/b84dcc7a-bafe-4610-95d2-1da1dc8700b4" />

Use same colors than number of books on shelves:

<img width="163" height="47" alt="Image" src="https://github.com/user-attachments/assets/a8a816bd-a3be-408f-9a39-6a97812dc9cb" />

(without https://github.com/openzim/gutenberg/issues/592 obviously)

PR #596, opened 2026-09-22
FIX: Show number of author books in book details view  #593

Abdul-Azhar-Jamesh:
issue  #593
Update BookDetailInfo.vue

Display the number of books available for the author in the current archive.
- Show the count beside the author name on the book details page.
- Reuse the existing localized `author.bookCount` message and shelf-count styling.


benoit74 (commented):
Rather than fetching the list of authors to find details about current book author, shouldn't we rather simply add the details we miss in book details?

@VikramAditya33 do you recall the rationale for not having `Author` class contains all `AuthorPreview` details in `scraper/src/gutenberg2zim/core/schemas.py`? I feel it is odd that a "Preview" object has more info than the full version.

VikramAditya33:
@benoit74 My bad, I'm unable to recall this, otherwise this is quite confusing. We should make Author extend `AuthorPreview`.

benoit74 (changes requested):
@VikramAditya33 no worries, happens to me so often...

@Abdul-Azhar-Jamesh please make `Author` class extend `AuthorPreview` and populate all properties (even if we do not use them right now in the UI). Also add all new properties to TS type.

benoit74 (changes requested):
LGTM, please squash commits and rebase on `main`

benoit74 (approved):
LGTM

Issue #582, opened 2026-09-19
Indgest Wikibooks OPDS streams


kelson42:
I think they are available at (in a few languages):
https://ws-export.wmcloud.org/opds/te/

Welcome page of the tool is at https://ws-export.wmcloud.org

Source code of the tool is https://gitlab.wikimedia.org/toolforge-repos/wsexport

Sriram-PR:
Started on this but hit a wall. ws-export.wmcloud.org is behind Anubis (the anti-bot proof-of-work thing) now - every path including /opds/ just returns the "Making sure you're not a bot" challenge page instead of the atom feed, so a normal http client can't fetch it. Checked a few times, it's not a temporary rate limit.

Only way around it would be solving the PoW in the scraper or driving a headless browser, both of which feel brittle and basically untestable in CI.

Is there a supported way to reach ws-export for a tool like this (allowlisted UA, token, a different host)? Or should we approach Wikibooks some other way given the wall? I can implement the opds ingestion once there's a reliable way to pull the feeds.

benoit74:
I don't see any Anubis on the ready-to-use XML exports, curl achieve to fetch them perfectily fine

Sriram-PR:
Yeah you're right, ignore what I said about Anubis. I was hitting /opds/ and /opds/en/ which just give the directory index, so I assumed it was walled off. /opds/en/Ready_for_export.xml pulls fine with curl, epub export too. My mistake, thanks.

PR #597, opened 2026-09-23
feat: add Wikisource as a scraping source (#582)

Sriram-PR:
Adds Wikisource as a third source (`--source=wikisource`, alias `WS`).

The issue title says Wikibooks, but the feeds it links (ws-export `/opds/`) are Wikisource. ws-export is the Wikisource export tool and doesn't cover Wikibooks at all, so this implements Wikisource. Can rename if you actually meant Wikibooks, but that would be a separate data source.

It mirrors the opentextbooks source. The catalog reads the per-language `/opds/<lang>/` index, finds the "ready for export" feed and parses its OPDS entries; each entry already carries title, author, language, license, year and per-format download links, so there's no separate metadata fetch. The pipeline downloads the epub from ws-export, validates it, pulls a cover and adds it to the ZIM.

Wikisource is per-language, so `--languages` is required. EPUB is the format that ships today (every "ready for export" book has one); pdf/html are marked unsupported per book, like the other sources handle missing formats.

One thing worth flagging: ws-export returns 403 for the default python-requests User-Agent (Wikimedia's UA policy), so the download engine now sends a descriptive UA with a contact URL. Small change in core, but it's correct for any Wikimedia endpoint and doesn't change the existing sources.

Tested against the live feeds end to end (discovers books, downloads valid epubs, builds a ZIM, works for non-latin languages like Telugu).

Refs #582


scraper/src/gutenberg2zim/sources/registry.py
```diff
@@ -110,6 +114,30 @@ class SourceProfile:
...
 )
 
+WIKISOURCE_PROFILE = SourceProfile(
+    slug="wikisource",
+    aliases=("WS",),
+    locale_namespace="wikisource",
+    display_name="Wikisource",
+    source_creator="wikisource.org",
+    zim_tags="_category:wikisource;wikisource",
+    zim_name_prefix="wikisource",
```
benoit74: I would prefer we use `wikisource-export` as `zim_name_prefix` so that we differentiate them more easily from Wikisource ZIMs we already have, created with mwoffliner scraper directly from Mediawiki itself.

README.md
```diff
@@ -118,13 +118,26 @@ The Open Textbook Library-specific options are:
 
 `--otl-ids` and `--books` cannot be used together.
 
+### Wikisource
+
+Select Wikisource with `--source=wikisource` (or `--source=WS`).
+
+Wikisource is organised per language, so `--languages` is required: the scraper reads the ["ready for export" OPDS feeds](https://ws-export.wmcloud.org/opds/) published by [ws-export](https://ws-export.wmcloud.org) for each requested language and downloads each book as an EPUB.
```
benoit74: Probably worth to clarify where one can find the list of supported languages codes (https://ws-export.wmcloud.org/opds/)

benoit74 (changes requested):
@Sriram-PR thank you very much for this first version of Wikisource. I feel like it works sufficiently well to deserve to be merged once we've solved inline comments + we've updated `offliner-definitions.json` to add the new source.

I've also identified a number of other bugs in the `core`, so I will open dedicated issues since these are not wikisource source responsibility.

scraper/src/gutenberg2zim/sources/registry.py
Sriram-PR: Got it, changed it to wikisource-export.

README.md
Sriram-PR: Added a line pointing to the ws-export OPDS index as the source of supported language codes.

Sriram-PR:
Added the Wikisource source to `offliner-definition.json` too (no source-specific flags, so its dependents list is empty). Thanks for the review.

VikramAditya33:
@Sriram-PR This is quite a substantial piece of work, and it involves UI changes as well. We’d prefer it if you could include screenshots of the UI in the PR description. This also helps us verify that the changes have been properly tested from your end before we review them.

Hope you understand!

scraper/src/gutenberg2zim/sources/wikisource/catalog.py
```diff
@@ -0,0 +1,199 @@
...
+ws-export (https://ws-export.wmcloud.org) publishes, per Wikisource language, an
+OPDS Atom feed of the books flagged "ready for export". Each feed lives under
+``/opds/<lang>/`` with a language-specific filename, so discovery reads the
+directory index to find the feed(s), then parses their ``<entry>`` elements.
+
+Every entry is self-contained: it carries the title, author, language, license,
+year, source page and one ready-made ws-export download link per format. So a
+`WorkRef` here already holds everything the metadata port needs - no per-book
+fetch, unlike Open Textbook Library.
+"""
```
VikramAditya33: Cover handling is incomplete. The selected book does not advertise an OPDS cover, and its EPUB declares `title.xhtml` as its cover. The shared extractor treats that XHTML file as an image and fails. More importantly, the English feed currently advertises explicit OPDS cover images for 74 books, but the catalog ignores every `rel="http://opds-spec.org/image"` link. The source should capture that URL and use it as a fallback when EPUB extraction fails. Books with no real cover should use the normal placeholder.

scraper/src/gutenberg2zim/sources/wikisource/metadata.py
```diff
@@ -0,0 +1,70 @@
+"""Wikisource metadata access.
+
+The OPDS feed entries parsed by `sources.wikisource.catalog` already carry every
+field the pipeline needs, including the per-format ws-export download links, so
+this port only maps a `WorkRef` into a `Work` - there is no per-book fetch.
+"""
+
```
VikramAditya33: The OPDS `rights` value is a URI, and `Work.license` is displayed verbatim. This currently renders `http://creativecommons.org/licenses/by-sa/3.0` in book details.
Please normalize known Creative Commons URIs to a human-readable label such as `Creative Commons Attribution-ShareAlike 3.0 (CC BY-SA 3.0)`. Ideally the label and canonical HTTPS URL would be represented separately so the UI can render a proper link.

locales/en.json
```diff
@@ -343,5 +343,39 @@
...
+  },
+  "wikisource": {
+    "about": {
+      "attribution": "Text provided by Wikisource, the free library.",
+      "introHeading": "About Wikisource",
+      "introParagraph1": "Wikisource is a free online library of source texts that anyone can improve.",
+      "introParagraph2": "Its volunteers transcribe and proofread books, documents, and other published works.",
+      "introParagraph3": "The texts are drawn from public-domain and freely licensed sources.",
+      "introParagraph4": "This library makes those texts available for offline reading."
+    },
```
VikramAditya33: `AboutView` unconditionally renders `missionHeading`, `missionAuthor`, and `missionParagraph1` through `missionParagraph17`, but none of those keys exist in the Wikisource namespace or common. They consequently appear literally in the UI.

scraper/src/gutenberg2zim/sources/registry.py
```diff
@@ -110,6 +114,30 @@ class SourceProfile:
...
+    catalog=WikisourceCatalog,
+    cli_options=wikisource_cli.CLI_OPTIONS,
+    parse_cli_options=wikisource_cli.parse_options,
+    handle_cli_action=wikisource_cli.handle_cli_action,
+    pipeline_options=lambda _mirror_url, _cache_dir: {},
+    metadata_options=lambda _cache_dir: {},
+    metadata_class=WikisourceMetadata,
+    pipeline_class=WikisourcePipeline,
+)
+
```
VikramAditya33: Wikisource works never populate `Work.collections`, but this profile advertises Collections. Since collection navigation is always enabled, the resulting page has an empty sidebar and largely duplicates the Books page.

locales/en.json
```diff
@@ -343,5 +343,39 @@
...
+    "metadata_defaults": {
+      "title": "Wikisource Library",
+      "description": "Free source texts from Wikisource"
+    },
+    "nav": {
+      "collections": "Collections"
+    },
+    "home": {
+      "allCollections": "All collections",
+      "mostPopular": "Most popular",
```
VikramAditya33: Wikisource does not assign `Work.popularity`, so every preview receives popularity 0. The UI therefore features the first OPDS entry while calling it “Most popular.”

scraper/src/gutenberg2zim/sources/wikisource/catalog.py
```diff
@@ -0,0 +1,199 @@
...
+        feed_urls = [
+            urljoin(index_url, href)
+            for anchor in soup.find_all("a", href=True)
+            if isinstance(anchor, Tag)
+            and (href := str(anchor["href"])).lower().endswith(".xml")
+        ]
+        if not feed_urls:
+            logger.warning("No OPDS feed found under %s", index_url)
+        return feed_urls
+
```
VikramAditya33: `_feed_urls()` accepts every `.xml` link in the directory. The directory currently contains only `Ready_for_export.xml`, but adding another feed would silently ingest unrelated entries and could produce duplicate works.
Please target the ready-for-export feed explicitly, validate the feed identity, or deduplicate discovered refs by work ID.

scraper/src/gutenberg2zim/sources/wikisource/catalog.py
```diff
@@ -0,0 +1,199 @@
...
+WIKISOURCE_SOURCE = "wikisource"
+BASE_URL = "https://ws-export.wmcloud.org"
+
+# OPDS acquisition media type -> the format name used across the pipeline.
+FORMAT_BY_MEDIA_TYPE = {
+    "application/epub+zip": "epub",
+    "application/x-mobipocket-ebook": "mobi",
+    "application/xhtml+xml": "xhtml",
+}
+
```
VikramAditya33: `mobi` and `xhtml` are stored on each `Work`, but the scraper only accepts `epub`, `html`, and `pdf`, so these formats can never be selected or exported.
Since this implementation currently supports only EPUB, it would be clearer to retain only EPUB acquisition links. Alternatively, XHTML needs to be mapped to and implemented as a complete offline HTML edition.

scraper/src/gutenberg2zim/sources/wikisource/metadata.py
```diff
@@ -0,0 +1,70 @@
...
+            languages=[language] if language else [],
+            license=extra.get("license"),
+            formats=[
+                Format(name=name, media_type=media_type, url=url)
+                for name, media_type, url in extra.get("formats", [])
+            ],
+            source_url=extra.get("source_url"),
+            extra={
+                "issued": extra.get("issued"),
+                "wikisource_page": extra.get("page"),
```
VikramAditya33: `issued` is parsed and carried in `Work.extra`, but it is never exported or displayed.

scraper/src/gutenberg2zim/sources/wikisource/metadata.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __init__(
+        self,
+        base_url: str = BASE_URL,
+        *,
+        engine: DownloadEngine,
+    ):
+        self._base_url = base_url.rstrip("/")
+        self._engine = engine
+
```
VikramAditya33: `base_url` and `engine` are stored but never used because all metadata comes from `WorkRef.extra`.

scraper/tests/test_wikisource_pipeline.py
```diff
@@ -0,0 +1,110 @@
+"""Tests for Wikisource per-work processing."""
+
```
VikramAditya33: No Wikisource integration test exists. The integration workflow still builds only Gutenberg and OTL. Wikisource needs a one-book ZIM in both `Tests.yaml` and `DailyTests.yaml`, including `zimcheck` and assertions for source metadata and EPUB presence.

VikramAditya33 (changes requested):
I've left some of the reviews, take a look at them. Another thing is Wikisource has a suitable popularity signal through MediaWiki’s `prop=pageviews` API. We can query ready-for-export page titles in batches, sum their `pageviews` over the last 60 days, and assign that value to `Work.popularity`. Core will then derive flames from the same metric, keeping ranking and flame display consistent.
This would also make the “Most popular” label accurate. Please avoid the top-pages endpoint because it only returns the top 1,000 pages and misses most ready-for-export works. The `pageview` request should be best-effort, use the scraper User-Agent, and avoid one HTTP request per work. @benoit74 Also might have some thoughts on this.

VikramAditya33:
@benoit74 I just saw that you created standalone issues to do this incrementally and this kinda duplicate my findings. @Sriram-PR You can ignore these for now and proceed with how @benoit74 prefers.

VikramAditya33:
@benoit74 I'll let you decide whether to go with my findings or not or better as separate issues.

Sriram-PR:
Thanks for splitting these into separate issues @benoit74, that makes them much easier to work through.

I'm good doing this incrementally. If you're happy merging this first version, I'll take the Wikisource follow-ups from there. I'd start on covers (#602) and descriptions (#601), and pick up the about-page text (#606) after the core simplified about-page change (#605) since it depends on it. Let me know if you'd rather I fold any of those in here before merging instead.

On @VikramAditya33's popularity idea: summing pageviews over the last ~60 days for the ready-for-export titles sounds like the right signal, and batching the queries keeps it to a handful of requests rather than one per book. I'd wire it up with the scraper UA and keep it best-effort so a pageviews hiccup can't fail the whole run. Fine either here or as its own issue, whatever you prefer.

benoit74:
@Sriram-PR @VikramAditya33 I indeed feel like incremental improvements is going to be easier to handle. Both in terms of implementation and review. I will open dedicated issues for valid concerns raised by @VikramAditya33 that I've missed.

benoit74:
> pick up the about-page text (https://github.com/openzim/gutenberg/issues/606) after the core simplified about-page change (https://github.com/openzim/gutenberg/issues/605) since it depends on it.

Nope, there is no dependency. #605 is about supporting the fact that wikisource does not have mission details in the about page. #606 is about adding mission details. So you can implement #606 now and then we do not need to implement #605 (even if I feel like it would be wise to have it implemented so that it could be possible for next sources we will implement)

scraper/src/gutenberg2zim/sources/wikisource/metadata.py
benoit74: Moved to https://github.com/openzim/gutenberg/issues/607

scraper/src/gutenberg2zim/sources/wikisource/catalog.py
benoit74: Moved to https://github.com/openzim/gutenberg/issues/608

locales/en.json
benoit74: Already tracked in #605 and #606 (only one of them needs to be implemented for now, even if implementing both is fine)

scraper/src/gutenberg2zim/sources/registry.py
benoit74: https://github.com/openzim/gutenberg/issues/609

locales/en.json
benoit74: See https://github.com/openzim/gutenberg/issues/610

scraper/src/gutenberg2zim/sources/wikisource/catalog.py
benoit74: See https://github.com/openzim/gutenberg/issues/611

scraper/src/gutenberg2zim/sources/wikisource/catalog.py
benoit74: See https://github.com/openzim/gutenberg/issues/612 and https://github.com/openzim/gutenberg/issues/613

scraper/src/gutenberg2zim/sources/wikisource/metadata.py
benoit74: See https://github.com/openzim/gutenberg/issues/614

scraper/src/gutenberg2zim/sources/wikisource/metadata.py
benoit74: See https://github.com/openzim/gutenberg/issues/615

scraper/tests/test_wikisource_pipeline.py
benoit74: See https://github.com/openzim/gutenberg/issues/616

